PR #4, opened 2018-11-01
Design system content updates

emilywaggoner:
Adding context to components: live examples, research, design decisions

### DIFF
--- a/src/_design-system/components/additional-info.md
+++ b/src/_design-system/components/additional-info.md
@@ -15,3 +15,11 @@ AdditionalInfo components are interactive components that reveal and hide conten
 ```html
 {% include  react/AdditionalInfo.njk %}
 ```
+
+## Examples
+
+You can see the Additional Info component on the Health Care Application form:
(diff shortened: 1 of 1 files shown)


jaredcunha-usds:
Would we if we'd need to show the components in the wild, especially on components where someone would need to authenticate or perform some interaction. I think showing them in context is nice, but I'm wondering about how we address these instance, especially at scale. 

- What happens when a component is removed from the page? How would we keep on top of this?
- What if a component example does not comport to usage guidelines (or gets modified in a way that doesn't)
- What if the components are specific to authenticated users?

emilywaggoner:
Yes I think this kind of context would require curation unless there is a way to pull back in live examples from the wild. Alternatively, we could use screenshots or animated gifs like the CMS design system does to demonstrate its responsiveness: https://design.cms.gov/guidelines/responsive/

jaredcunha-usds:
Would it be helpful to show the component with something other than dummy text?

PR #16, opened 2018-11-30
Some notes from Beth for a couple entries

bethpotts:
Here's a start at comments for you on the style guide.


jaredcunha-usds:
@jenniferlee-dsva Do you want to review this or should I go ahead and merge it in?

jenniferlee-dsva:
@jaredcunha-usds  Hi Jared - I want to touch base with Beth first, so please leave it as is now. Thank you for checking!

PR #30, opened 2019-03-11
Update va-forms.md

jenniferlee-dsva:
Hi @jaredcunha-usds - I changed the name of this file from "VA forms" to "Government forms." 

1. Could you make sure the label that displays in the nav also shows "Government forms"? 

2. Is there any way to edit the width of the first column, so that the form name/number appear all on one line, rather than wrapping?  It would actually be preferable to wrap the long text name, but better to show the form number name all on one line.


jaredcunha-usds:
@jenniferlee-dsva The only way to do something like that would be to inline a CSS width property, which unless you're able to edit the HTML directly, that would not be possible. There's sort of a hack you can do, which is adding non-breaking spaces to the longest row header, which will extend the column in all instances. I went ahead and did that, which you can see [here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/30/commits/61a3992875e3b0919a7c1e6cf1d7d2841c9ac320#diff-fde9066d79bdd340f9f4758ceb5ec197R44). 

The non-breaking space is replacing all spaces with `&nbsp;`. 

I've also renamed the file to `government-forms.md` so it properly appears in the left rail alphabetically. Let me know if you want me to merge and deploy this.

jenniferlee-dsva:
Oh, just figured out, I have to comment first, then merge. Thx!

PR #33, opened 2019-03-18
WiP - Prod and staging

omgitsbillryan:


jaredcunha-usds:
Is this one ready to merge or did you want to hold for any reason?

omgitsbillryan:
@jaredcunha-usds ya, sorry, please let me merge this one. It's in conjunction with some PR's in another repo.

wyattwalter:
I think we can just use a single tarball called "site.tar.bz2" or "build.tar.bz2", whichever makes sense. Since there's nothing environment-specific at this point, we can just use one and promote through the envs.

jaredcunha-usds:
@omgitsbillryan - I think it might be helpful to see separate staging env so we can pass around some documentation while still a work in progress. Doesn't have to be on the first pass, but it would be nice to have as some designers were wondering how they can preview content updates.

PR #39, opened 2019-03-27
Adding we, us, you text

jenniferlee-dsva:
@jaredcunha-usds - just checking if I used your table html snippet correctly. Does it look ok?

### DIFF
--- a/src/_content-style-guide/we-us-you.md
+++ b/src/_content-style-guide/we-us-you.md
@@ -13,7 +13,9 @@ When referring to VA, use the first-person pronoun (we, us, our). Use the second
 <div class="do-dont__do">
 <h3 class="do-dont__heading">Like this</h3>
 <div class="do-dont__content" markdown="1">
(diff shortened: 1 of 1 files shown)


jaredcunha-usds:
@jenniferlee-dsva  Looks good to me. If you want in the future, you can even add some context to the boxes as well. 

Screen Shot 2019-03-27 at 6 17 13 PM

I'll merge this in tomorrow so this message doesn't disappear on you and you don't see my reply.

jenniferlee-dsva:
Great suggestion - hard to tell without seeing it closer to what it actually looks like. 

Please remove the quotation marks around the example though. And then it's perfect!

jaredcunha-usds:
@jenniferlee-dsva I just did the example in the browser, I didn't update the code. If you'd like, you can update this pull request to incorporate or leave as is for now. Update later too. Up to you.

jenniferlee-dsva:
ah, now I get it. okay. let me try...

PR #40, opened 2019-03-27
Update and rename lists.md to bulleted-lists.md

jenniferlee-dsva:
@jaredcunha-usds -- Added contents and changed filename to "bulleted-lists" to reflect topic. Would like this to appear in the nav as Bulleted lists. Not sure if just changing the filename will do that. Can you check and deploy if no issues?

### DIFF
--- a/src/_content-style-guide/bulleted-lists.md
+++ b/src/_content-style-guide/bulleted-lists.md
@@ -1,0 +1,42 @@
+---
+layout: default
+title: Bulleted lists
+---
+
+# Bulleted lists
+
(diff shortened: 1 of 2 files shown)


src/_content-style-guide/bulleted-lists.md
```diff
@@ -0,0 +1,42 @@
+---
+layout: default
+title: Bulleted lists
```
jaredcunha-usds: @jenniferlee-dsva This is the field that determines how a page will appear in the left nav AND browser tab title

PR #42, opened 2019-03-28
Danielle edits, response, additions

DanielleSoCompany:
Hi @jenniferlee-dsva. I've addressed all of the questions, with the exception of a couple which I'm asking Beth and Randi to look at. Lots of edits as well as some notes and questions. Feel free to reach out with any questions. Thanks!


jenniferlee-dsva:
I reviewed Danielle's patch-3 edits and incorporated edits; removed acronyms to be moved to its own topic page; and removed resolved items. There are still some questions that need to be addressed, but these should be easier to spot now.

PR #46, opened 2019-04-09
Update seo.md

jenniferlee-dsva:
@jaredcunha-usds 
This is the MD page that I used your snippet for inserting a screen image. Could you check to make sure it looks okay before you deploy? 

I'll also attach the screenshot file here: 
browser-title-example-va-prescription-refill-and-tracking


jaredcunha-usds:
@jenniferlee-dsva I've pushed a minor change here. I've uploaded an image, but I retook the screenshot because the one you attached had gotten compressed and appeared blurry (or it was taken from a 1x dpi screen). 

I did a minor correction on the image tag as well. The tag itself was correct, but you wrapped in the "```", which in markdown, is to write code. Which is what I did for the get the styling on the example.

Let me know if this needs to be merged.

jenniferlee-dsva:
Thanks @jaredcunha-usds - I like the changes to the screenshot example. One thing though - in the screenshot you Slacked, I see a couple of the examples above that screenshot example and they have a weird table around them.   Those shouldn't have tables around them, but not sure how I can fix them. I think the error might be generated by the fact that we use the pipe | Veterans Affairs in title tags, and MD is interpreting that pipe as a table element.

PR #47, opened 2019-04-10
Update and rename health-content to health-content.md

jenniferlee-dsva:
@jaredcunha-usds - I added this new MD. Could you make sure it looks okay? I used your Like this/Not this snippet, but hard to tell if the formatting looks okay without a real preview.

### DIFF
--- a/src/_content-style-guide/health-content.md
+++ b/src/_content-style-guide/health-content.md
@@ -188,7 +188,7 @@ Our National Center for PTSD is the world's leader in PTSD research, education, 
 </div>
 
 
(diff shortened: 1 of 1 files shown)


jaredcunha-usds:
@jenniferlee-dsva I pushed a fix to this page, one the boxes was missing a closing tag. 

I had also changed each of the numbers to level 3 headings. The advantage to this is that an anchor link appears when you hover on them, which makes it easier to share specific sections on that page.

Since the domain is now live, let's make sure to keep this as a pull request until internal questions are resolved. (where you have a note for Danielle at the bottom, for example)

Here's a screenshot of how the page looks.

Screenshot_2019-04-11 Health content - VA gov Design System

PR #55, opened 2019-04-19
WiP - Make links pretty using :permalink

omgitsbillryan:
### Background
When serving through jekyll or github pages, there's native support for "pretty links" (e.g. GET `/components/accordions` will point serve up `/components/accordions.html`). S3 website hosting does not have this feature. [The initial solution](https://github.com/department-of-veterans-affairs/devops/pull/4468) was for the deploy process to remove the .html extension from all files before uploading to S3, and then set the files' `content-type` metadata as `application/html`. Ultimately, this solution is turning out to be more complex than desired ([see new PR here](https://github.com/department-of-veterans-affairs/devops/pull/4521)). I'm exploring the use of `permalinks` as an alternate solution so that the `_site/` is in a more "ready-to-go" form without the need of file manipulation at deploy time.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -44,26 +44,26 @@ sass:
 collections:
   documentation:
     output: true
-    permalink: /:collection/:title
+    permalink: /:collection/:title/
     label: Documentation
   content-style-guide:
     output: true
-    permalink: /:collection/:title
+    permalink: /:collection/:title/
   design:
     output: true
-    permalink: /:collection/:title
+    permalink: /:collection/:title/
   components:
     output: true
-    permalink: /:collection/:title
+    permalink: /:collection/:title/
   patterns:
     output: true
-    permalink: /:collection/:title
+    permalink: /:collection/:title/
   layout:
     output: true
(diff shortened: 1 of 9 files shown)


omgitsbillryan:
@jaredcunha-usds do you think this is a reasonable approach that will scale with this application?

jaredcunha-usds:
I do tend to prefer URLs without the trailing slash. Is that going to be problematic?

I noticed two issues on this PR where there are 404s 
1) All the overview pages when you click Overview in the left nav
2) All the responsive previews 
Screen Shot 2019-04-19 at 12 25 30 PM

omgitsbillryan:
I too prefer the pretty links. For now I'm trying to weight the options - are the pretty links worth a complex deployment process? Is there some other way, maybe different from what I've done in this PR, that we can maintain those pretty links but also have the assets in a "ready-to-go" format?

I haven't fully hashed out all the broken links in this PR. My goal for this PR was mostly to start a conversation.

omgitsbillryan:
Doesnt look like we're going this route.

PR #69, opened 2019-04-29
WIP: all or none of the above

jaredcunha-usds:
This PR will codify rules to observe when designing forms with checkboxes/radios/selects where all or none of the options may apply.

What I want to know:
- Is anything on this page incorrect?
- Would there be anything helpful that I could add?
- Should we keep as its own page or make a page with other form patterns (though this could become a huge page)
- Any objections to documenting this?

Screenshot:
Screenshot_2019-05-01 None or all of the above - VA gov Design System


goldenmeanie:
I think this is helpful! A few comments:
1. > When it comes to designing forms, as a general rule, simpler is usually better.

    could be:

    > As a general rule simpler is usually better. In forms this can sometimes mean breaking complicated questions into a larger number of easy to understand decisions.
2. I like the simplification for checkboxes (no need for "none of these"). The one thing that I'd check in practice is whether there are any business rules about needing an explicit negative for policy reasons. In those cases it's probably better to reframe as a yes/no followed by additional clarification checkboxes as required. This would also help in a "required field" situation. I think this might make sense as an aside here to address a common question. 
3. I think it would be a small improvement to align the (please check all that apply) copy in the explanation with the copy in the visual example.

src/_patterns/none-or-all-of-these.md
```diff
@@ -0,0 +1,38 @@
...
+
+When it comes to designing forms, as a general rule, simpler is usually better.
+
+## Checkboxes
+
+In the case where the response is a set of checkboxes, below the question, specify that the user may "check all that apply" or that they "may check more than one."
+
+<img src="{{site.baseurl}}/images/more-than-one-checkbox.png" alt="set of checkboxes" style="max-width: 376px">
+
+**Do not** include checkboxes for "all of these" or "none of these" since the user may already select as many or as few boxes as necessary; implied by the text: "check all that apply." Introducing these options could potentially introduce unnecessary complexity to interaction design and the codebase, and perhaps even to data models.
```
rtwell: I can see how “None of these apply” could be enticing because a user may look at this list and say “none of these apply to me, but I have to check _something_ but what if my selection has a negative impact on the outcome of this form?”. I think the way to address this is to clarify within the question and its accompanying text. Using the example Jared posted, were a user of middle-eastern decent, none of those categories apply to that person. It would be helpful to say something like “If you do not identify with any of the categories listed here, please chose ‘White’”. The full question might be:

**Which categories best describe you?**
You may check more than one. If you do not identify with any of the categories listed below, please select “White”.

rtwell: @goldenmeanie i saw you commented in the issue, sorry to make us type in two places!

jaredcunha-usds:
@goldenmeanie 

1. Yes. Love that. 

2. I don’t think there is a current business need because this pattern documentation is a response to a pull request we didn’t accept. But yeah, I think if the data is required, you can make the checkboxes into a follow-up of another question. 

3. It’s not directly. I’m hoping that maybe someone can confirm, but perhaps that would have been in our forms already if people were getting hung up.

4. Not 100% sure what you mean here. Can you clarify?

5. Planning on it!!!

@rtwell yeah, this might be a bad example. I think the question should be phrase differently and we should not be forcing people to identify as something they are not nor imply they are not significant enough to include. Is there a better sample we can use?

goldenmeanie:
1. Yay!
2. _I think I am not 100% tracking here_. Required data of this type is a scenario I've seen in a few places. I'm proposing noting that this is a thing to check out in the docs. 
3. As Ryan says, "none of the above" is a really common pattern that people expect and in high stakes scenarios I think we may be introducing anxiety by simplifying it out. It might be better to violate the form designer logic (using mutually exclusive checkboxes) to provide a more supportive environment. 

    This is a strong statement, and if we can't point to a specific study, it might be better not to mention research. This item seems like maybe a conversation topic for Wednesday?
4. I didn't read closely enough. The second example "You may check more than one" is already matching, I got thrown off by the first quoted example copy "Check all that apply" being different than the visual pattern example. Maybe flip order to show matching copy first? But this is clearly mostly user error on my part.
5. Yay!

jaredcunha-usds:
Cool. This is helpful. 

I was unaware that we might have set of checkboxes out there already where the user is required to select at least one option, and that option can include "none of these." But if that is the case, maybe we don't document this at all and we should have accepted the [BAH work](https://github.com/department-of-veterans-affairs/vets-website/pull/9944) on this? I assumed they introduced it because it has not come up before.

But if that's not the case…

In a high stakes scenario, I think we might either need to write the questions in a different way (ex: Would you like to provide information about your race? [yes/no] | if yes, a follow up where they select). Otherwise, we could be more assertive by letting them know they don't have to make a selection. 

I'm going to hold on merging this until next week and maybe we can hash this out a little further.

jaredcunha-usds:
Closing this. Moving the page with updated screenshot here:
https://dev-design.va.gov/patterns/forms

PR #70, opened 2019-04-29
WIP: changes to background-only alerts

jaredcunha-usds:
based on our conversation in the design, trying to clean up documentation on alert component, specific, the background-color-only modifier. 

For details, see this comment: https://github.com/department-of-veterans-affairs/va.gov-team/issues/52#issuecomment-487728461

### DIFF
--- a/src/_components/alertboxes.md
+++ b/src/_components/alertboxes.md
@@ -20,7 +20,7 @@ anchors:
 
 ## Information status
 
-Used to provide helpful information to a user.
+Used to provide helpful information to a user or something that warrants attention. Not used for negative consequences.
 
 <div class="site-c-showcase">
 {% include_relative html/alert-informational.html %}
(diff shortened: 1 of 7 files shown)


jenniferlee-dsva:
Hi @jaredcunha-usds - I'm not sure what exactly is being changed with this merge. I'm still concerned that the guidance around how the warning and informational alerts should be used in static/non-application content; and also that the nomenclature is still confusing -- those alerts are not really used to convey "status" at all (the way the background-only components are.) 

Can I work with you to incorporate the guidance and clarification we reached from the first design meeting? Also re: background alerts audit - I think @kevwalsh content anomaly modeling has already identified them (at least in the non-authenticated pages), so you wouldn't need to duplicate that for non-auth content.

jaredcunha-usds:
@jenniferlee-dsva You can see the changes in the [files tab](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/70/files). 

The documentation on this page comes mostly from USWDS. We didn't have any documentation of our own, so I just grabbed what came from the original source of which our design system is based. It seems we may have deviated from that somewhat as we use them for content and we need to adapt our our documentation.

Happy to work with you tomorrow if you have time.

jaredcunha-usds:
@jenniferlee-dsva and I reviewed the content updates on this page. I going to go ahead and merge this into master so it can be viewed on dev-design.va.gov (on the VA network only ATM). I will deploy later in the afternoon if I don't hear anything.

PR #72, opened 2019-05-02
full width alert documentation

jaredcunha-usds:
The full-width alerts are now in Formation. Here's some documentation for them. 
Screen Shot 2019-05-02 at 5 25 41 PM

One thing I want us to do is to stop calling these "banners," so I am not going to reference them as such. They are a variant on the alert component, so we should call them full-width alerts, as the code implies, but I'm open to "page-width alerts" too, as long as we call it some kind of alert - though I think we should name them as we do in the code so we all use the same language.  

Please review the usage in the screenshot or page. I think I referenced everything in the github issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,12 +13,20 @@
       }
     },
     "@department-of-veterans-affairs/formation": {
-      "version": "6.2.3",
-      "resolved": "https://registry.npmjs.org/@department-of-veterans-affairs/formation/-/formation-6.2.3.tgz",
-      "integrity": "sha512-gcY+LqaIrgc31lvQ0Sn3kg7Yevq6joK1qnVFN2cpNeqosHRElSxrvDA+UjFBspBxW0mbXLTXME7iyZWfGtpwdw==",
+      "version": "file:../veteran-facing-services-tools/packages/formation",
       "requires": {
-        "@fortawesome/fontawesome-free": "5.7.2",
+        "@fortawesome/fontawesome-free": "5.8.1",
         "foundation-sites": "5.5.3"
+      },
+      "dependencies": {
+        "@fortawesome/fontawesome-free": {
+          "version": "5.8.1",
+          "bundled": true
+        },
+        "foundation-sites": {
+          "version": "5.5.3",
+          "bundled": true
+        }
       }
     },
     "@department-of-veterans-affairs/formation-react": {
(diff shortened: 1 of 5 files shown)


jaredcunha-usds:
Are you talking about on the alert or the body content?

If alert - that's the alert CSS as it is. 

If body content - that's the scaffolding we have. That might be a bit of work because that's something that affects every page on design.va.gov.

rtwell:
image

that's how i was hoping it would look

also, i _think_ we need to make dismissable an option? that functionality may be specifically for the drupal version, but im not sure

jaredcunha-usds:
I think some design fixes might be needed, but that's not quite in scope for what this is for. I moved the CSS as it was on `vets-website` into `formation` yesterday. There's a lot that can be probably cleaned up in that, but that wasn't what I wanted to include because it completely changes the acceptance criteria from _does this still work as it was_ to _does this still work as it was, does it support updated specs, does the close functionality work_.  Also, that work cannot be done on this repo anyway.

The pull request here is about documentation.

rtwell:
ohhhhh gotcha, thanks! 👍

jaredcunha-usds:
Is the documentation okay? @jenniferlee-dsva?

jenniferlee-dsva:
Thanks Jared. This captures all the general guidance. I'd like to make some of it a little more explicit for our VA stakeholders who will use these.  Can we add: 


•	Emergency homepage alerts notify Veterans, VA employees, and the public of events that affect VA services or site features. 
•	To ensure that customers always know they can find critical service information in this area, do not use emergency homepage alerts for general press, outreach, or administrative messages. 

Might there be a way to show an example combo message, so people have a sense of what the guidance means? We provided OPIA template text for that kind of situation, as that has happend in the past. (See # 2 message on this page: https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Banners/homepage-emergency-banners-preapproved-020919.md )

jaredcunha-usds:
@jenniferlee-dsva Might be hard to tell, but I incorporated your feedback.

Screen Shot 2019-05-03 at 4 00 40 PM

jaredcunha-usds:
By the way, this might be a good candidate to add to the styleguide perhaps? I can help you if you want to show the examples in actual alerts… if this sounds like a good idea to you.

jenniferlee-dsva:
I see it! Nit picky thing - the "Some VA facilities and services are affected" should be the actual alert heading, no? (Otherwise it looks like you have a heading, and then the body text is another heading in bold.)  Other than that, it looks good!

jenniferlee-dsva:
Alerts are a component type, so I think the guidance is more useful in context of the component itself.

jaredcunha-usds:
@jenniferlee-dsva I think the heading in that type of alert might be something like "Hurricane X updates" or something that to that extent. I'm trying to keep the examples on the generic, so I can use something like "Sample subheading" as well if you feel more comfortable with that.

PR #87, opened 2019-05-20
DRAFT page titles

jenniferlee-dsva:
@DanielleSoCompany @bethpotts CC @RLHecht  - I’m unhiding the style guide topic for “Page titles and section titles” in the content style guide site. Please take a look at this PR page and let me know if you see any issues. 

Right now, we don’t have any character count limit per se for page titles and section titles ( H1s and h2, h3 headers). We do have  a count for the whole meta browser title in the SEO meta guidelines however. 

@DanielleSoCompany  - If we work backward from the SEO guidelines, it would mean that the character limit for H1 page titles are 41 with spaces (because 60 is the max with | Veterans Affairs).  

I was looking around our pages and some of the page title H1s are more like 50-60 characters long.  

I don’t see a big problem with that length - 60 characters for page titles. I think even searches accommodate a little longer than our recommended 60 total for the meta title. 

Would you all be okay with allowing 60 characters for H1s -- and alternately,  revising the guidance for SEO to allow longer browser title?  If we allow 60 max for H1s, then, the max for meta browser tile would be 60 + space+ | Veterans Affairs - 79.

### DIFF
--- a/src/_content-style-guide/page-titles-and-section-titles.md
+++ b/src/_content-style-guide/page-titles-and-section-titles.md
@@ -35,7 +35,7 @@ Section and subsection titles (also sometimes called headers and subheads) help 
 
 - Try to keep section and subsection titles to 70 characters maximum, with spaces. 
 
-We allow a little more character count for sections and subsections than page titles. But in general, sections become hard to scan when they're too long.  Eliminate unnecessary details or nuance in section and subsection titles, and address them with more depth in the paragraph copy.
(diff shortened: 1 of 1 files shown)


jenniferlee-dsva:
@jaredcunha-usds  - For some reason, I cannot add Danielle Thierry as one of the assignees to review. Would you be able to add her for me? I've been able to with other content style guide PRs, so it's kind of odd.

jaredcunha-usds:
@jenniferlee-dsva I just added her as a collaborator, so you should be able to add her now in the future.

jenniferlee-dsva:
Merging to go live

PR #102, opened 2019-06-05
update additional info docs

jaredcunha-usds:
@rtwell  Made a couple of minor tweaks to the documentation. Let me know what you think of these. See below for inline notes.


src/_patterns/additional-info.md
```diff
@@ -5,20 +5,24 @@ draft: true
 ---
 # Additional info
 
-## Use case
-Sometimes, an interaction on a benefit application or resource is so nuanced that it has labels or requirements that some users don't fully understand.
+<p class="va-introtext">Sometimes, an input or information is so nuanced that it has labels or requirements that some users don't fully understand.</p>
```
jaredcunha-usds: Removed references to specific locations of additional info so that designers do not feel limited about the pattern's usage outside benefit applications. 

The H2 here was probably unnecessary here, and also, an accessibility issue because there was no content under the H1.

src/_patterns/additional-info.md
```diff
@@ -5,20 +5,24 @@ draft: true
...
 
 ## What is Additional info
-The additional info pattern is a context specific interaction where we provide users with relevant information that helps gain about the decision they are about to make. This information is hidden beneath a quasi-accordion-like interaction because it isn’t necessary for the majority of users and would impede the primary action on the page. 
+The additional info pattern is a context specific interaction where we provide users with relevant information that helps gain about the decision they are about to make. This information is hidden beneath a quasi-accordion-like interaction because it isn’t necessary for the majority of users and would impede the primary action on the page.
 
-## Examples
+## Example
+
+The following example depicts how to use additional info to describe a medical condition if the user is unsure what to call it.
```
jaredcunha-usds: - A single example, so I removed the pluralization from the heading. 
- Added brief description of the example
- Added some captions to describe the images.

src/_patterns/additional-info.md
```diff
@@ -27,8 +31,21 @@ When the user clicks, they see an explanation of what to search for, and an exam
 ## Pattern specifics
 The [current component of the Additional info](https://design.va.gov/components/additional-info) interaction has a dotted line underneath it, along with a down-carat to indicate interactivity
 
-### Do: 
-Help users by providing contextual, relevant information that brings clarity to content they are taking action. 
-
-### Don’t:
-Don’t use this pattern to house information that is critical to the form or content relevant to the specific interaction a users has to take.
```
jaredcunha-usds: Here I just replaced the headings with the do/don't pattern

src/_patterns/additional-info.md
```diff
@@ -27,8 +31,21 @@ When the user clicks, they see an explanation of what to search for, and an exam
...
+</div>
+</div>
+<div class="do-dont__dont">
+<h3 class="do-dont__heading">Don’t</h3>
+<div class="do-dont__content" markdown="1">
+Don’t use this pattern to house information that is critical to the form or content
+</div>
+</div>
+</div>
+
```
jaredcunha-usds: I think it might be useful to explain why we don't use accordions here

jaredcunha-usds:
I feel a little uneasy about the name of this pattern because it conflicts with a component name and we are trying not to blur any lines between components and patterns. How about:

- Added context
- Supplementary info
- Supplemental instruction

Or I guess any combo of these words?

src/_patterns/additional-info.md
rtwell: love it. done.

src/_patterns/additional-info.md
rtwell: you are my hero

src/_patterns/additional-info.md
rtwell: it probably would be if i knew the rationale 😬 

if i had to guess i would say that it's not an accordion because the content is not whole unto itself and needs the "parent" content which makes it relevant. an accordion is a chunk of content that is self-contained.

jaredcunha-usds: I believe I've heard @goldenmeanie or @lalexanderson-dsva relay something to that extent. Can one of you confirm?

goldenmeanie:
I propose `contextual help`

jaredcunha-usds:
@goldenmeanie What about something to like this? I was thinking perhaps we name the "pattern" supplementary info. This is generic enough to where we can accommodate a few different facets:
- Using the Additional Info component for **contextual help**, as Ryan has documented. I like the term _contextual help_ here because it is really descriptive of what is provided… help, in context. Because it's hidden behind an interaction, I think it can be considered _supplementary_. 
- Information we'd put at the bottom of a page, like related links.  I think it can be considered _supplementary_.

Thoughts?

goldenmeanie:
@jaredcunha-usds It sounds like you are talking about two subsection on this page that would cover different components and scenarios? If so, that makes sense to me. Though I wonder if it means we should assimilate the accordion in here too?

Also, I was midstream on some edits that may appear as if I ignored your comment. That is not the case!

src/_patterns/additional-info.md
```diff
@@ -3,30 +3,31 @@ layout: default
 title: Additional info
 draft: true
 ---
-# Additional info
+# Contextual help
 
-<p class="va-introtext">Sometimes, an input or information is so nuanced that it has labels or requirements that some users don't fully understand.</p>
+<p class="va-introtext">In cases where a Veteran input can have large or complicated impact on outcomes we use the contextual help to locate expanded guidance next to the relevant part of the form.</p>
```
jaredcunha-usds: I think we might be using the term _user_ instead of _Veteran_ in the documentation since the user will not always be the Veteran.

src/_patterns/additional-info.md
```diff
@@ -35,17 +36,24 @@ The [current component of the Additional info](https://design.va.gov/components/
...
+
 </div>
 </div>
 </div>
 
 ## Why not use an Accordion component?
 
-Accordions, due to their visual presentation, can be disruptive to the way that users navigate through a form or read dense content on the page. Think of using the [additional info component]({{site.baseurl}}/components/additional-info) to provide a speed bump whereas an accordion would provide a stop sign.
\ No newline at end of file
+Accordions, are designed for static content and are overwhelming to insert in between inputs. Think of using the [additional info component]({{site.baseurl}}/components/additional-info) to provide a speed bump whereas an accordion would provide a stop sign.
```
jaredcunha-usds: For the note on static content… is that going to be clear to new designers? By static, do you mean content that does not come from the CMS, or do you mean page body content?

jaredcunha-usds:
@goldenmeanie Yeah, I think for some of this pattern documentation, we want to think about how we can handle subsections to consolidate like things, as wel do for Benefit Applications, form design, etc.

Contextual help seems to be used both in and outside of forms… something @rtwell may want to capture?  (this can be done later)

I added some comments on your edits BTW.

goldenmeanie:
> Contextual help seems to be used both in and outside of forms… something @rtwell may want to capture? (this can be done later)

I consider these rogue instances that should be updated to something else

src/_patterns/additional-info.md
goldenmeanie: Can you clarify what you mean by "here?" My intent is to make a distinction between forms/tools (contextual help) and static content (accordions). The in-form reasoning is above, but the static content constraint is to preserve the value of the pattern and not conflict with text links, which happen a lot more in content pages than forms.

src/_patterns/additional-info.md
goldenmeanie: Good question, I need better shorthand. By static content I mean "content outside of a form or tool"

src/_patterns/additional-info.md
jaredcunha-usds: @goldenmeanie - Sorry, missed this comment yesterday. By "here" I meant in this place in the page. The github UI hid the text I was talking about, but it's this, which you've already provided some edits:
> Accordions, are designed for static content (outside forms or tools) and are overwhelming to insert in between inputs. Think of using the [additional info component]({{site.baseurl}}/components/additional-info) to provide a speed bump whereas an accordion would provide a stop sign.

I'm satisfied with this explanation, so just closing this loop.

PR #120, opened 2019-06-21
Appeals additions to Word List

RLHecht:
Adding the appeals entries to word list as well as the terms Blue Water and Brown Water.


RLHecht:
@jenniferlee-dsva This is ready for you to look at.

PR #135, opened 2019-08-06
content updates/cleanup

peggygannon:


jaredcunha-usds:
@peggygannon I'll go ahead an merge this in. Just an FYI, You don't need to use <br> tags in markdown if you make use like breaks, as you would if you were typing paragraphs in Word. This can be a separate commit, but would you be to remove the content you have in brackets, for example, instances of _[screenshot]_?

PR #137, opened 2019-08-19
Fix link to Sketch file on home page

emilywaggoner:
The current link at the bottom to the Sketch file is broken, I've updated it with the correct URL.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -92,7 +92,7 @@ homepage: true
         </a>
       </div>
       <div class="vads-l-col--12 medium-screen:vads-l-col--6 vads-u-padding--2">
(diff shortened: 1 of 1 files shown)


jaredcunha-usds:
Hey @emilywaggoner, just getting back from vacation, but I must I have missed that link when updating. I think we should use the link on the design.va.gov domain instead of GitHub: 
https://design.va.gov/downloads/VA-gov-Pattern-Library.sketch

jaredcunha-usds:
resolved in this pr: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/138

PR #164, opened 2020-01-03
Update button names to align with CSS class names and documentation

andaleliz:
I updated all button names so that:

- Primary is now called Default
- Secondary is now called Primary
- Tertiary is now called Secondary

This aligns with the guidance on [design.va.gov/components/buttons](https://design.va.gov/components/buttons), and class names.  I believe this will reduce confusion, especially for new designers.

One thing I noticed is that the guidance/docs gives examples of a `Big` version of all three buttons, but the library file only has it available in the blue (default) style.  Should we have the other two versions available as big too?


andaleliz:
@CrystabelReiter thought this route may prevent the issues we had the last time we committed files to the repo. Tagging you in hopes you can review it. I also wanted to tag Emily, but can't seem to find her GH handle to assign it to her. I slacked her for awareness.

andaleliz:
@CrystabelReiter Thank you! Merged the PR, and will add the big buttons separately.

PR #175, opened 2020-03-03
Adding entry on street addresses

RLHecht:
Adding an entry in the dates, times, and phone numbers section of style guide to include guidance on street addresses.


jenniferlee-dsva:
@RLHecht - Could you review/check my changes?

PR #181, opened 2020-04-02
April What's New topics added

RLHecht:
Added list of style guide changes for April in the What's New section.


RLHecht:
@jenniferlee-dsva I fixed the MD code so the html shouldn't be showing anymore. I was missing </p> after each section.

jenniferlee-dsva:
Thx @RLHecht

PR #191, opened 2020-06-09
Adding TTY and accessibility info for phone numbers

RLHecht:
Adding to style guide that TTY numbers should always be linked. 

Also adding in the source code that needs to be written for both phone numbers and TTY numbers so that screen readers can read them correctly.

### DIFF
--- a/src/_content-style-guide/dates-and-numbers.md
+++ b/src/_content-style-guide/dates-and-numbers.md
@@ -86,13 +86,21 @@ from 8:00 a.m. to noon CT
 
 ## Phone numbers 
 
-Use hyphens between numbers, and don't use parentheses to set off the area code: 212-123-1234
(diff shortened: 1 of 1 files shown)


RLHecht:
@jenniferlee-dsva I made your edits from above. Let me know if they are approved and I can merge. Thanks!

DanielleThierryUSDSVA:
Hi @RLHecht. 2 quick things:

1) "Regular phone numbers and TTY numbers should always be hyperlinked" - Could we change to this: "Hyperlink all phone numbers, including TTY numbers." so we shift to the active voice.
2) I looked at the 2nd commit, and the change didn't look like it quite matched Jen's edit above so just wanted to bring to your attention. I may be missing something, but thought I'd raise to you for a 2nd look.

Thanks!

PR #193, opened 2020-06-16
adding +1 inside aria label code for phone numbers

RLHecht:
@DanielleThierryUSDSVA I updated this entry per our convo yesterday about keeping the +1 in the code if it's not causing any issues.


RLHecht:
@DanielleThierryUSDSVA Just found this. Can you please review and merge? Thank you!

PR #286, opened 2020-08-13
Fixing typos in footer

andaleliz:
The VA311 number in the footer had an extra 8 in it and a `1-` before the rest of the phone number. Removed all of that to match what we see on the site footer on production.


andaleliz:
I did pull down the latest version and update the library before I made these changes.  I thought it was a little weird that within the library file it notified me that updates to the library were available, but ultimately I thought it made sense since there are components in the library that use symbols from the library.

To be extra safe, you could pull down my file and try it on your end to confirm it's all good.

I just realized forgot to update the change log, oops! I can close this, update and make a new PR if you'd like.

CrystabelReiter:
Pulled the file down and it looks good to me! Thanks for catching it ! I can add that typo to the next update's changelog if you don't feel like adding it.

andaleliz:
Cool, that sounds good! It's not a major change so I think it's okay to leave it out of the log for now

PR #287, opened 2020-08-18
Update messaging-dictionary.md

jenniferlee-dsva:
The scheduled system maintenance messages are pointing to the wrong component (should reference Ryan's updated banner design, not the promo announcement banner);  also, the link is going to 404.

Ryan's new sitewide downtime banner design is here in Formation: https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/visual-design/components/maintenancebanner/  but it's missing the UX guidance, which I believe still needs to be added to the components section of the design system. We had a ticket for that - I'm sorry I can't find it anymore.


pdavies88:
Owner and reviewer no longer with VA teams. If code needs to still be fixed please open with a new PR.

pdavies88:
Owner and reviewer no longer with VA teams. If code needs to still be fixed please open with a new PR.

PR #313, opened 2020-09-15
Update bulleted-lists.md

jenniferlee-dsva:
One of the LIKE THIS boxes look out of whack - super skinny. Copy/pasted html to see if it will resolve.

image


RLHecht:
@jenniferlee-dsva These aren't showing weird for me on my screen. What browser are you using?
Screen Shot 2020-09-15 at 3 30 01 PM

KevinMHoffmanUSDS:
@jenniferlee-dsva We were checking on this ticket in the DS Council today, is this still an issue? If so we can look into it.

humancompanion-usds:
I'm not able to reproduce and this change added a space which would not fix the reported issue. Closing.

PR #321, opened 2020-10-08
Remove comma after state names

RLHecht:
@DanielleThierryUSDSVA or @bethpotts  Could you please review this PR and merge? We had commas after the state abbreviations in the Like this Not this boxes, which shouldn't be there.


DanielleThierryUSDSVA (approved):
Looks good to me, @RLHecht!

PR #333, opened 2020-10-23
Delete deploy_site.sh

cvalarida:
This script was used to deploy to GitHub pages, but is no longer called.
image

Now, we deploy to S3 (see the [Jenkinsfile](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/master/Jenkinsfile#L36-L51)).


bkjohnson (approved):
For further evidence, we don't even have a `gh-pages` branch anymore.

PR #334, opened 2020-10-27
Revert recent merges

cvalarida:
I'm not sure why, but one of these changes broke design.va.gov/components (and a few others). When tested locally, I couldn't reproduce the failure, so I'm reverting two recent merges: [Delete `deploy_site.sh`](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/commit/6f184fcfb42fff73fee77b7dc680639979ae2d76) and [Update Jekyll](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/commit/0767ed7fcf81a4af017046293ce3838d1c1263cf).


deploy_site.sh
```diff
@@ -0,0 +1,11 @@
+rm -rf _site
```
bkjohnson, line 1: I doubt removing this script caused the problem, but :shrug:

bkjohnson (approved):

deploy_site.sh
cvalarida: :thinking: Interesting.

PR #348, opened 2020-11-24
Circleci deploy

bkjohnson:
Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/15169

in order to keep the deploy from overwriting the storybook deploy that is happening from the `veteran-facing-services-tools` repo, we need to do this in Circle rather than Jenkins.

Once this is merged, I will delete the following YAML configs from the devops repo:

- `deploys/vets-design-system-documentation-vagov-dev.yml`
- `deploys/vets-design-system-documentation-vagov-staging.yml`

We will only deploy to dev and staging when a new commit gets merged into master:

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
       - run: bundle install
       - run: yarn build
       - run: bundle exec jekyll build
-      - run: aws s3 sync _site s3://<< parameters.bucket >>
+      - run: aws s3 sync _site s3://<< parameters.bucket >> --acl public-read
 
 # Orchestrate or schedule a set of jobs
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,7 @@ jobs:
 
     steps:
       - checkout
+      - aws-cli/setup
       - run: yarn install
         # If this works, think about adding it to the Gemfile
       - run: gem install bundler:1.16.6
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,9 @@ jobs:
       - image: cimg/ruby:2.7.2-node
     environment:
       AWS_DEFAULT_REGION: us-gov-west-1
+    parameters:
+      bucket:
+        type: string
 
     steps:
       - checkout
@@ -20,12 +23,13 @@ jobs:
       - run: bundle install
       - run: yarn build
       - run: bundle exec jekyll build
-      - run: aws s3 sync _site s3://dev-design.va.gov
+      - run: aws s3 sync _site s3://<< parameters.bucket >>
 
 # Orchestrate or schedule a set of jobs
 workflows:
-   deploy:
+   deploy-dev:
     jobs:
       - build-deploy:
+          bucket: "dev-design.va.gov"
           context:
             - Platform
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,8 +2,6 @@
 version: 2.1
 # Use a package of configuration called an orb.
 orbs:
-  # Declare a dependency on the welcome-orb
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -2,12 +2,40 @@
...
+
+    steps:
+      - checkout
+      - aws-cli/setup
+      - run: yarn install
+        # If this works, think about adding it to the Gemfile
+      - run: gem install bundler:1.16.6
+      - run: bundle install
+      - run: yarn build
+      - run: bundle exec jekyll build
```
bkjohnson, line 25: This creates the static site files in the `_site` directory.

.circleci/config.yml
```diff
@@ -2,12 +2,40 @@
...
+      - run: aws s3 sync _site s3://<< parameters.bucket >> --acl public-read
```
bkjohnson, line 26: This will deploy the static site to the bucket

.circleci/config.yml
```diff
@@ -2,12 +2,46 @@
...
+  deploy-dev:
     jobs:
-      - welcome/run
+      - build-deploy:
+          bucket: "dev-design.va.gov"
+          filters:
+            branches:
+              only: master
+          context:
+            - Platform
```
bkjohnson, line 38: The `context` is where the environment variables that allow us to use the `aws` cli are coming from.

.circleci/config.yml
```diff
@@ -2,12 +2,46 @@
...
-  welcome:
-    # Run the welcome/run job in its own container
+  deploy-dev:
     jobs:
-      - welcome/run
+      - build-deploy:
+          bucket: "dev-design.va.gov"
+          filters:
+            branches:
+              only: master
```
bkjohnson, line 36: `bucket` is a parameter defined by the job, and `filters` is a filter that will only run this workflow when we are on the `master` branch

Jenkinsfile
```diff
@@ -50,25 +50,5 @@ pipeline {
       }
     }
 
-    stage('Deploy dev & staging') {
```
bkjohnson, line 53: This is removed since we will be doing it from `CircleCI`.

.circleci/config.yml
```diff
@@ -2,12 +2,46 @@
...
+      AWS_DEFAULT_REGION: us-gov-west-1
+    parameters:
+      bucket:
+        type: string
+
+    steps:
+      - checkout
+      - aws-cli/setup
+      - run: yarn install
+        # If this works, think about adding it to the Gemfile
```
cvalarida, line 21: Any follow up action items here?

.circleci/config.yml
cvalarida: Is this worth adding as a comment in the file?

.circleci/config.yml
cvalarida: Is this worth adding as a comment in the file?

cvalarida (approved):
This looks good to me, but I'd like to get ops eyes on it.

@dginther can you give it a quick double-check? The only kind of CI configuration I'm familiar with is GitLab's. This is close enough for me to be comfortable approving, but I'm not sure if I'm missing anything tricky.

.circleci/config.yml
bkjohnson: Not yet. I'd like to address it with a follow-up PR that will implement caching/optimizations once we get the CI for storybook working properly.

.circleci/config.yml
bkjohnson: I don't think we need a comment for `filters`, but I'll add one for `bucket` since it is a parameter.

.circleci/config.yml
bkjohnson: Possibly. I'll add it.

jbritt1 (approved):

PR #354, opened 2020-12-08
Add storybook link

bkjohnson:
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/16119

This is the change to the "Components" page that will be visible:

image

~~If we decide to move the production deploy from Jenkins to CircleCI before this gets merged, then we can get rid of the `Jenkinsfile` changes and remove the check for the environment variable.~~

Now that #358 has been merged, if we want this to only appear in some environments we will have to make some changes to the CircleCI config.


Jenkinsfile
```diff
@@ -26,10 +26,10 @@ pipeline {
       }
     }
 
-    stage('Build static site') {
+    stage('Build static site for production') {
       steps {
         sh 'npm run build'
-        sh 'bundle exec jekyll build'
+        sh 'JEKYLL_ENV=production bundle exec jekyll build'
```
bkjohnson, line 32: From [Jekyll's docs](https://jekyllrb.com/docs/configuration/environments/):

> The default value for `JEKYLL_ENV` is `development`.

By setting this to `production` for the Jenkins build, we can use the `jekyll.environment` variable to conditionally render things in our markdown files.

src/_components/index.md
```diff
@@ -4,6 +4,17 @@ title: Components
...
+  <div class="feature">
+    <h3>Come check out our <a href="{{ site.baseurl }}/storybook" class="vads-u-color--link-default vads-u-font-weight--bold vads-u-text-decoration--none">Storybook pilot</a>!</h3>
+    <p>We are eager to receive feedback from designers & developers on their opionions on Storybook. Here are some example questions:</p>
+    <ul>
+      <li>Can you find guidance on how to use a component?</li>
+      <li>Are you able to see how a component will appear in a mobile viewport?</li>
+    </ul>
+  </div>
+{% endif %}
+
```
bkjohnson, line 17: Using the environment variable, this will appear in any non-production build.

cvalarida (approved):
Looks good! Is this "part of" that ticket or does it close that ticket?

cvalarida:
@peggygannon Do you have any thoughts on voice / tone / anything else?

**Context:** We're adding this blurb on https://design.va.gov/components/ to point folks to https://design.va.gov/storybook/ as an alternative solution for component documentation. After a while, we'll be sunsetting https://design.va.gov/components/ in favor of Storybook.

We're also adding a link to https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/visual-design in [this PR](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/555) and will be removing that section after a while in favor of Storybook.

bkjohnson:
I think the reason I used "Part of" instead of "Closes" is that back then I wasn't sure if we were going to do things differently by environment. I'll update it now.

PR #369, opened 2021-01-07
Deploy fix

bkjohnson:
This fixes a recently noticed problem where we had `*.html` files from recent deploys, but older extensionless html files that were taking precedence in the browser.

Context: https://dsva.slack.com/archives/G01BJ3ESXL4/p1610037389032700?thread_ts=1610037383.032600&cid=G01BJ3ESXL4

Our solution is modeled after: https://simpleit.rocks/ruby/jekyll/tutorials/having-pretty-urls-in-a-jekyll-website-hosted-in-amazon-s3/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,9 @@ jobs:
       - run: bundle install
       - run: yarn build
       - run: bundle exec jekyll build
+        # We are taking these extra steps due tosome differences between Jekyll and AWS S3.
+        # More context:
+        #   https://simpleit.rocks/ruby/jekyll/tutorials/having-pretty-urls-in-a-jekyll-website-hosted-in-amazon-s3/
       - run:
           name: Remove .html extension on non-index files
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,13 +25,27 @@ jobs:
       - run: bundle exec jekyll build
       - run:
           name: Remove .html extension on non-index files
-          command: find _site/ -type f ! -iname 'index.html' -iname '*.html' -print0 | while read -d $'\0' f; do mv "$f" "${f%.html}"; done
+          command: |
+            find _site/ -type f ! -iname 'index.html' -iname '*.html' \
+            -print0 | while read -d $'\0' f; do mv "$f" "${f%.html}"; done
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -23,7 +23,32 @@ jobs:
       - run: bundle install
       - run: yarn build
       - run: bundle exec jekyll build
-      - run: aws s3 sync _site s3://<< parameters.bucket >> --acl public-read
+        # We are taking these extra steps due tosome differences between Jekyll and AWS S3.
```
cvalarida, line 26:
```suggestion
        # We are taking these extra steps due to some differences between Jekyll and AWS S3.
```

.circleci/config.yml
```diff
@@ -23,7 +23,32 @@ jobs:
...
+        # More context:
+        #   https://simpleit.rocks/ruby/jekyll/tutorials/having-pretty-urls-in-a-jekyll-website-hosted-in-amazon-s3/
```
cvalarida, line 28: :+1:

To save our future selves in the event that this link breaks, I'd also recommend a short explanation. Something like "To access a .html file served from S3, the URL needs to have the .html extension. We're removing the extension to make the URLs prettier."

.circleci/config.yml
```diff
@@ -23,7 +23,32 @@ jobs:
...
+            -print0 | while read -d $'\0' f; do mv "$f" "${f%.html}"; done
+      - run:
+          name: Sync extensionless html files with correct type
+          command: |
+            aws s3 sync _site s3://<< parameters.bucket >> \
+            --acl public-read \
+            --delete \
+            --exclude "storybook/*" \
+            --exclude "*.*" \
+            --content-type "text/html"
```
cvalarida, line 42: Here's hoping all extensionless files end up being HTML! :crossed_fingers:

.circleci/config.yml
```diff
@@ -23,7 +23,32 @@ jobs:
...
+      - run:
+          name: Sync remaning files
```
cvalarida, line 44:
```suggestion
          name: Sync remaining files
```

cvalarida (approved):

.circleci/config.yml
bkjohnson: image

PR #371, opened 2021-01-11
Make sure storybook doesn't get deleted

bkjohnson:
A [recent change to the deploy pipeline](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/369) introduced a bug which accidentally deleted the `/storybook` directory. This fixes it to prevent that from happening.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,9 +60,9 @@ workflows:
       - build-deploy:
           # Set the bucket parameter to be the dev bucket
           bucket: "dev-design.va.gov"
-          filters:
-            branches:
-              only: master
+          # filters:
+          #   branches:
+          #     only: master
           context:
             # The Platform context has the environment variables for setting up the aws cli
             - Platform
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -48,9 +48,9 @@ jobs:
             aws s3 sync _site s3://<< parameters.bucket >> \
             --acl public-read \
             --delete \
-            --exclude "storybook/*" \
             --exclude "*" \
-            --include "*.*"
+            --include "*.*" \
+            --exclude "storybook/*" \
```
bkjohnson, line 53: Order matters!

cvalarida: Fun fact! :sweat_smile: 

Also, prolly shouldn't have that trailing `\`. I'm surprised it works with it there.
```suggestion
            --exclude "storybook/*"
```

cvalarida (approved):

PR #377, opened 2021-01-21
Remove hardcoded HTML extension from links

bkjohnson:
@mchelen-gov noticed a bug where a link was hardcoded to a page with an `html` extension, but the S3 bucket only has html files without an extension. This fixes that issue, but there are lots of other examples of this same thing happening.

## Testing

Double checked that the links worked locally

## Screenshots

After these changes, these are the only things that match my regex, and I _think_ they're ok:

image

[Slack thread for additional context](https://dsva.slack.com/archives/CBU0KDSB1/p1611250163022100)


cvalarida (approved):
Ahh, interesting.

PR #416, opened 2021-03-31
update link section and add ruby version to readme

k80bowman:
closes https://app.zenhub.com/workspaces/vsp-design-system-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/va.gov-team/21571

Updates link section to look like this:
image

### DIFF
--- a/src/_design/html/links.html
+++ b/src/_design/html/links.html
@@ -1,6 +1,6 @@
 <p>This is a <a href="javascript:void(0);">default text link</a></p>
 
-<p>This is an <a href="javascript:void(0);">active text link</a></p>
+<p>This is an <a style="font-weight:700" href="javascript:void(0);">active text link</a></p>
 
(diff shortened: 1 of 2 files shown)


src/_design/html/links.html
```diff
@@ -1,8 +1,10 @@
-<p><a href="javascript:void(0);">This</a> is a text link on a light background.</p>
+<p>This is a <a href="javascript:void(0);">default text link</a></p>
 
-<p><a class="usa-color-text-visited" href="javascript:void(0);">This</a> is a visited link.</p>
+<p>This is an <a href="javascript:void(0);">active text link</a></p>
```
bkjohnson, line 3: There appears to be no styling marking this as active. Can that be added?

image

k80bowman: I got so focused on the visited link that I forgot to come back to that. Yes, thank you for calling that out.

k80bowman:
updated, should now look like this: 
image

bkjohnson (approved):

PR #422, opened 2021-04-07
Updating time zone info

RLHecht:
Our style is to use only ET with hours. This needs to be added to the rules about phone numbers and the examples need to be updated. We also need to add directions about always including hours and days of operation when listing a phone number.


bethpottsVADEPO (commented):
Randi, let's not drop the article "the" before "office"--can you add it?

PR #433, opened 2021-04-16
Using iframes to Storybook for component documentation

cvalarida:
Like it says on the tin, this PR is a proof-of-concept for using iframes pointing to Storybook stories. They'll look like this:
image

Usage is pretty straightforward:
```
{% include storybook-preview.html story="components-back-to-top--default" %}
```
We can even include an optional `height` parameter in the `include` statement if desired.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
       - run: gem install bundler:1.16.6
       - run: bundle install
       - run: yarn build
-      - run: bundle exec jekyll build
+      - run: JEKYLL_ENV=production bundle exec jekyll build
         # We are taking these extra steps due to some differences between Jekyll and AWS S3.
         # To access a .html file served from S3, the URL needs to have the .html extension.
         # We're removing the extension to make the URLs prettier.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,8 @@ jobs:
     parameters:
       bucket:
         type: string
+      envConfig:
+        type: string
 
     steps:
       - checkout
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -22,7 +24,7 @@ jobs:
       - run: gem install bundler:1.16.6
       - run: bundle install
       - run: yarn build
-      - run: bundle exec jekyll build
+      - run: bundle exec jekyll build --config _config.yml,<< parameters.envConfig >>
```
cvalarida, line 27: According to [Jekyll's documentation](https://jekyllrb.com/docs/configuration/options/), this is how we can use a shared config and override individual values with others. In this case, we're using one config for all environments and then overriding some values with environment-specific configs.

I opted to go this route so we can set environment-specific variables such as the environment type (which we may later use to determine how to gather analytics, for example).

src/_includes/storybook-preview.html
```diff
@@ -0,0 +1,5 @@
+<div class="site-showcase">
+  <iframe width="100%" height="{{ include.height }}" src="{{ site.storybook_path }}/iframe.html?id={{ include.story }}&viewMode=story"></iframe>
+</div>
+
+[See this in Storybook]({{ site.storybook_path }}/?path=/docs/{{ include.story }})
```
cvalarida, line 5: Should we link to the docs page or the story page? :thinking: I was about to change this to `/story/`, but thought maybe it might make more sense to link to the prop docs? :man_shrugging: I'm not opinionated either way, I think.

src/_components/action-links.md
```diff
@@ -6,9 +6,7 @@ title: Action links
 
 # Action links
 
-![action links]({{site.baseurl}}/images/action-links-updated.png) 
-
-[Interact with this component on Storybook](https://design.va.gov/storybook/?path=/docs/components-action-link--page)
+{% include storybook-preview.html height=500 story="components-action-link--page" %}
```
bkjohnson, line 9: Can we make the height bigger to get rid of this scroll bar?

image

bkjohnson (approved):
Seems like a good proof of concept.

.circleci/config.yml
cvalarida: Yeah. I figure that's outside the scope of this PR, though, so I didn't make the switch. I expect adding parameters to jobs isn't going to be a problem when switching to GH actions, so I'm not worried about this adding something that we won't be able to do later.

Issue #267, opened 2020-07-17
Heading typography is different on design.va.gov from the production site


1Copenut:
# Feature Request
@jenstrickland brought up a great point in a Slack thread yesterday about the difference in heading typography between design.va.gov and production:

> There's a difference in heading styling as noted in the VA design system and Formation CSS.
>  * In the design system, H2, H3, H4, and H6 are Source Sans Pro. H1 and H5 are Bitter.
>  * In Formation, all headings are Bitter.
> 
> In general, serif fonts may be more difficult for readers to comprehend, especially for long, bold text strings. If the design system council has an opportunity to revisit the history of the difference in the VA design system documentation and the Formation CSS, that would be helpful. The sans-serifs will be easier, quicker to read for the most people. I started to document in an issue ticket, but then wasn't sure which issue template was the correct one to use.

---
![Screen Shot 2020-07-14 at 10 32 13 AM](https://user-images.githubusercontent.com/934879/87829753-425e5b00-c845-11ea-8a36-bf5f71495b43.png)

---
![87439104-5e040000-c5be-11ea-8bc1-eacd77d03ce2](https://user-images.githubusercontent.com/934879/87829763-47bba580-c845-11ea-8355-314b2a6a163d.png)

- [ ] I’ve searched for any related issues and avoided creating a duplicate issue.

## This update is for:

- [ ] Content styleguide
- [ ] Component
- [ ] Pattern
- [ ] Utility
- [ ] Other

## What is the name

This is a two-part issue. The first thing to do short-term is to update design.va.gov to match what our headings are doing in production.

The longer-term thing would be for the design system council to explore considering Source Sans Pro for headings. Some questions include whether this is all headings, all headings expect H1s, and what level of effort this might require. This would likely depend on a robust visual regression test suite; changing CSS always has some level of risk for visual breaks.

## What is the nature of this update?

Documentation and potentially production CSS

- [ ] How to build this component/pattern
- [ ] When to use this component/pattern 
- [x] When to use something else
- [x] Usability guidance
- [ ] Accessibility
- [x] Implementation
- [ ] Research insights
- [ ] Package information

1Copenut:
Pinging @KevinMHoffmanUSDS && @rtwell && @CrystabelReiter for awareness.

jenstrickland:
[...]

PR #447, opened 2021-05-04
Remove the header font family override

cvalarida:
Why were they even there in the first place? :confused:

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/267

**Note:** This doesn't update the actual headers on design.va.gov. They're still using Source Sans Pro. I don't know what the history of that is. :man_shrugging: 

## Screenshots
image
**Note:** The H6 is still using Source Sans Pro, but that's how it is on va.gov as well. :man_shrugging:

### DIFF
--- a/src/assets/stylesheets/_components/_showcase.scss
+++ b/src/assets/stylesheets/_components/_showcase.scss
@@ -6,12 +6,6 @@
   padding: $units-2;
 }
 
-.site-showcase > h2,
-.site-showcase > h3,
-.site-showcase > h4 {
-  font-family: $font-sans;
-}
-
 .site-showcase + .highlighter-rouge {
   margin-top: -1px;
 }


CrystabelReiter (approved):

src/assets/stylesheets/_components/_showcase.scss
```diff
@@ -6,12 +6,6 @@
   padding: $units-2;
 }
 
-.site-showcase > h2,
-.site-showcase > h3,
-.site-showcase > h4 {
-  font-family: $font-sans;
-}
-
```
bkjohnson, line 14: This looks like it doesn't remove the override for `<h1>` - is that in some other CSS file?

bkjohnson: Nevermind, it looks like the `<h1>` and `<h5>` in your screenshot _are_ `Bitter`.

Looks like here is where we would remove the `<h6>` override:

https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/c7e0380f4c37f5309bbe32ff227ed22350e09540/packages/formation/sass/base/_b-utils.scss#L61-L63

cvalarida: The h1 already has the correct font family (because it wasn't overridden with the incorrect one here). Don't ask me why. :man_shrugging:

bkjohnson (approved):

PR #452, opened 2021-05-05
Sort the side nav based on order

cvalarida:
With this change, we can add an `order` property to a page's front matter to change the order of that page in the side nav and the mobile nav menus.
image

### DIFF
--- a/src/_includes/_mobile-nav.html
+++ b/src/_includes/_mobile-nav.html
@@ -11,7 +11,8 @@
       <li class="site-mobile-nav-list__item">
         <a class="site-mobile-nav-list__link {% if page.url contains "documentation/index" %} current {% endif %}" href="{{ site.baseurl }}/documentation">Overview</a>
       </li>
-      {% for p in site.documentation %}
+      {% assign sortedDocumentation = site.documentation | sort: "order", "last" %}
+      {% for p in sortedDocumentation %}
        {% if p.layout == "default" %}
           {% unless p.index or p.draft %}
             <li class="site-mobile-nav-list__item">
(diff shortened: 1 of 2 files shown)


src/_includes/_mobile-nav.html
```diff
@@ -170,28 +177,29 @@
...
+    <button class="usa-accordion-button site-mobile-nav__accordion-button" aria-expanded="{% if page.url contains "/experimental-design/" %}true{% else %}false{% endif %}" aria-controls="nav-a8">
+      Experimental Design
+    </button>
+  </p>
+  <div id="nav-a8" class="usa-accordion-content site-mobile-nav__accordion-content" aria-hidden="{% if page.url contains "/exerimental-design/" %}false{% else %}true{% endif %}">
+    <ul class="site-mobile-nav-list">
+      <li class="site-mobile-nav-list__item">
+        <a class="site-mobile-nav-list__link {% if page.url contains "experimental-design/index" %} current {% endif %}" href="{{ site.baseurl }}/experimental-design">Overview</a>
+      </li>
+      {% assign sortedExperimentalDesign = site.experimental-design | sort: "order", "last" %}
```
bkjohnson, line 191: Sneaky sneaky change in with some indentation.

src/_includes/_mobile-nav.html
```diff
@@ -11,7 +11,8 @@
       <li class="site-mobile-nav-list__item">
         <a class="site-mobile-nav-list__link {% if page.url contains "documentation/index" %} current {% endif %}" href="{{ site.baseurl }}/documentation">Overview</a>
       </li>
-      {% for p in site.documentation %}
+      {% assign sortedDocumentation = site.documentation | sort: "order", "last" %}
```
bkjohnson, line 14: This is clever - I had to look it up how `sort` is working. The second argument (`"last"`) tells the [sort where to put things that don't have an `order` property](https://jekyllrb.com/docs/liquid/filters/#sort)

bkjohnson (approved):

src/_includes/_mobile-nav.html
cvalarida: Oh, good call. I should have linked to the filter's documentation. :+1:

PR #468, opened 2021-05-12
Instructions for adding a markdown file to experimental design

cvalarida:
We'll probably want to add to this when we finalize the process for _when_ to add a component or pattern to the experimental section. Also, we need an actual template. :sweat_smile: 
image


humancompanion-usds:
I pulled these changes in while I was working on https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/708 The /about/contributing-to-the-design-system page now includes these Github UI instructions.

PR #473, opened 2021-05-28
Update contributing-to-the-design-system.md

sshein:
I suggested a few changes to this. I also am still confused after reading this what the "rules" are for moving an experimental component to the official design system. After a team validates it once, then can it move over? And if so, who decides that and who moves it over?

### DIFF
--- a/src/_documentation/contributing-to-the-design-system.md
+++ b/src/_documentation/contributing-to-the-design-system.md
@@ -11,9 +11,9 @@ How to suggest a new component, update existing components, and retire component
 
 
 ## Suggest a component
-Anyone can suggest a new component. We currently keep a list of suggestions in the design system backlog. 
+Anyone can suggest a new component. We currently keep a list of suggestions in the design system backlog.   
 
-Follow these 5 steps to suggest a component for the design system.
+Follow these 6 steps to suggest a component for the design system.
 
 ### 1. Check the design system backlog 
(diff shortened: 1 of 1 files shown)


allison0034:
Is this missing step #2?

sshein:
good call somehow the step 2 header got removed! I'll add it back in

sshein:
@allison0034 does everything else look good? I'll probably just merge and then do a quick follow-up with any needed changes once all comments are made

src/_documentation/contributing-to-the-design-system.md
```diff
@@ -24,9 +24,11 @@ If you do not see your idea in the backlog at all go to step 2.
 Fill out a [Experimental Design System request ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/new?assignees=&labels=&template=experimental_design_request.md). You will need to provide justification for this new component and any artifacts you might want to include.
 
 ### 3. Present to Design System Council
-You will present your work to the Design System Council at an upcoming meeting.
+You will present your work to the Design System Council at an upcoming meeting. Experimental design requests will always be prioritized in this meeting, and teams can assume generally that they will be able to present the same week of the request.
 - Submit request to join an upcoming Design System Council meeting in #vsp-design-system 
 - During the meeting, the Design System Council will evaluate the request and make a decision.
```
mpelzsherman, line 29: Can we add more info here about how often the Design Council meets, and what happens if the proposal is rejected?

PR #475, opened 2021-06-01
initial codespaces config

mchelen-gov:


cvalarida (approved):
:tada: The specific bundler version is annoying. :roll_eyes: Eventually, we'll get around to updating bundler and the Gem.lock file, but it's pretty low on the priorities for now. :shrug:

This is great, thanks!

PR #479, opened 2021-06-11
Add search for experimental design section

cvalarida:
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/25447
The title says it all.
image

How it works: https://github.com/christian-fei/Simple-Jekyll-Search#create-searchjson

### DIFF
--- a/src/search.json
+++ b/src/search.json
@@ -45,6 +45,21 @@ layout: null
   {% endunless %}
   {% endfor %}
 
+  {% for item in site.experimental-design %}
+  {% unless item.layout == "iframe" %}
+    {
+      "title"    : "{{ item.title | strip_html | escape }}",
+      "section"  : "Experimental design",
+      "url"      : "{{ site.baseurl }}{{ item.url }}",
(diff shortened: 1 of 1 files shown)


src/search.json
```diff
@@ -45,6 +45,21 @@ layout: null
   {% endunless %}
   {% endfor %}
 
+  {% for item in site.experimental-design %}
```
bkjohnson, line 48: I realize that this is a json file, but it would be nice if we could merge `site.experimental-design`, `site.components`, etc. all into one array rather than having to copy and paste stuff and only change the array that we loop through. The `section` is different as well, but that could probably be handled easily enough.

I didn't find an immediately obvious way to do that in liquid, though :(

bkjohnson (approved):

src/search.json
cvalarida: Potentially there is a way, but :man_shrugging: This is probably _clearer_ than doing some Liquid magic.

PR #480, opened 2021-06-14
Update principles.md

sshein:
added a principal based on our discussion, Ryan - maybe you can look it over and make sure you think it's unique enough to include :)


src/_design/principles.md
```diff
@@ -24,6 +24,8 @@ VA.gov design principles are fundamental pieces of guidance that help designers
 
 **Respect Veterans’ time by using research to validate your best conceptual approach.** Rather than having Veterans spend time helping you resolve granular design decisions, focus your research on the primary user problem or need. 
 
+**Base your design decisions on whether the Veteran or caregiver can complete a task over what might "look right."** Everything we design should ultimately make tasks easier for our Veterans. Don't spend time finessing how something looks if it's working for the Veteran.
```
rtwell, line 27: I'm wondering if we should include something about using default design system components and patterns

PR #493, opened 2021-07-15
Remove postinstall docs

cvalarida:
As of wc-v0.7.1, this is no longer needed.


bkjohnson (approved):
Pre-emptively approving

PR #496, opened 2021-07-21
Adding quotation mark entry

RLHecht:
We need to add an entry about always using curly quote marks on VA.gov instead of the straight up and down ones.

### DIFF
--- a/src/_content-style-guide/00_word-list.md
+++ b/src/_content-style-guide/00_word-list.md
@@ -288,6 +288,8 @@ Job Bank: Neither word is plural. See Veterans Job Bank entry.
 
 lewisite: Define on first mention in body copy as "a natural compound that contains the poison arsenic." Include definition in another sentence, as a parenthetical, or set off with commas.
 
+LGBTQ+: Use in all instances.
+
 life insurance: Do not hyphenate in the form of a compound adjective: “life insurance coverage.”
 
 loan guaranty agent: Lowercase. Capitalize only when used as someone's title. The preferred style is to recast the sentence so that it adheres to AP Style on professional titles.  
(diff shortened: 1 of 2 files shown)


RLHecht:
@DanielleThierryUSDSVA I added the LGBTQ+ entry onto this PR

DanielleThierryUSDSVA:
Hi @RLHecht and @bethpottsVADEPO. A few things on this PR:

- **For the LGBTQ+ term:** I've added a bit more description for this one. I pulled the description from the LGBTQ+ health program page. Thoughts? I'd like to publish this one today.
**- For the curly quotes and apostrophes:** 
  -  I noticed that we have straight quotes/apostrophes in this style guide PR. Before we go telling people not to use them, we should fix all of the ones in the guide, yes? If this will take some time, can we break that out into a separate PR so we can publish the LBGTQ+ term PR?
  - I think we should add examples in each so it's clear what we're talking about. Like this:  
  Use curly apostrophes (ex. don’t) instead of straight up and down apostrophes (ex. don't). Thoughts?

RLHecht:
@DanielleThierryUSDSVA Your LGBTQ+ entry looks good to me. I agree that we should make sure the style guide has curly quotes throughout before we tell users that's what they need to do on VA.gov. I'll pull that out of this PR so that we can publish the LGBTQ+ change now.

RLHecht:
@DanielleThierryUSDSVA I added in the examples for both apostrophes and quotes. I just went through the style guide and I think there's a script to make all straight up and down marks curly. It shows wrong in the github file but it's correct on the site, so we don't need to go through all the pages to correct anything.  @kelsonic helped me mark up the quote marks that need to be straight up and down in our examples of what not to do.

PR #498, opened 2021-07-24
Add timestamps

cvalarida:
# Description
Optionally adds timestamps to the bottom of the page. They only appear if the `date_added` or `last_updated` frontmatter appear on the page. This is intended for use in the experimental design section to keep track of when an experiment was added.

Ideally, these timestamps are updated automatically, but that was more work to figure out. :man_shrugging: This is the quick and dirty way to get _some_ information out there.

# Screenshots 
image

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem "minima", "~> 2.0"
 # If you have any plugins, put them here!
 group :jekyll_plugins do
   gem "jekyll-feed", "~> 0.6"
+  gem "jekyll-last-modified-at"
 end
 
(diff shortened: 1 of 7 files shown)


bkjohnson (commented):
I'm a bit concerned about relying on manual updates to the `last_updated` value. I expect that this will get ignored/forgotten, which will lead to misleading documentation, and that kinda defeats the purpose of this.

cvalarida:
Agreed. How might we automate it?

bkjohnson (approved):
I added a plugin which allows us to automate the modified date by adding a `last_modified_at` property to page objects.

PR #499, opened 2021-07-27
Create expandable-alert.md

lildziner:
@sshein here's a first pass at the expandable alert. The [code snippet from a live Vet Center page](https://www.va.gov/springfield-ma-vet-center/) isn't working for me.

### DIFF
--- a/src/_experimental-design/expandable-alert.md
+++ b/src/_experimental-design/expandable-alert.md
@@ -1,0 +1,34 @@
+---
+layout: default
+sub_section: Alert boxes
+title: Mobile friendly expandable alert
+---
+Suggested by: Leyda Hughes, VSA Facilities Team
+
+# Purpose
+
+When an alert box is too visually prominent and additional information needs to be communicated without leaving the page.
+
+## Description
+Mobile friendly expandable alerts combine the Additional Info component within Background Color Only alert color schemes, and include an icon. 
+
+### Example:
+
+![image](https://user-images.githubusercontent.com/4960080/127147819-12d07cf2-b357-4058-81d5-56f0caf764a2.png)
+
+![image](https://user-images.githubusercontent.com/4960080/127148001-517c85d7-e6d9-401d-81d7-1d3b8f1dbeb1.png)
+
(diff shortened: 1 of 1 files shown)


sshein:
cool :) maybe @cvalarida can help with the code? you can always ping him on slack too. also someone told me Riley's team coded this in somewhere sharable, is that the case? (maybe ping Justin or Riley and ask?)

bkjohnson:
I did some digging and it looks like you're not seeing what you expect because `vets-website` has a script that is identifying that element and applying some additional transformations on it:

https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/static-pages/facilities/vet-center/createExpandableOperatingStatus.jsx

cvalarida:
@bkjohnson What can we do to make this work here?

lildziner:
@LinnJS I heard you have a better working example of the expandable alert on the debt portal. Is that something that you can point us to?

bkjohnson:
We might be able to extract the specific `static-pages` files, compile them down using webpack, and include that script bundle on this page.

Another option would be to rewrite it in a way that avoids React - either like we have done in `formation` or with a web component.

sshein:
if it seems too complicated, can we just include a screenshot of the component open and closed, and then a link out to wherever folks can find the code?

lildziner:
I don't think I have write access to on this repo to merge the pull request. I tried approving first but it won' let me do that either. Am I missing something obvious?
[screenshot of GH telling me I can't approve my own PR]

bkjohnson:
It won't let you approve it since you're the author.

I took a look at the settings and the `design.va.gov-writers` group has "write" permissions, meaning they should be able to merge PRs. @sshein is a maintainer of that group, so she should be able to give you access which will give you the permissions you need.

sshein:
ah ok, I can also just merge for you Leyda for now. Is it ready to be merged?

lildziner:
Yes, please @sshein can we try. We can refine from there?

sshein:
ok I merged it should be up in a bit, keep a look out :) https://design.va.gov/experimental-design/

sshein:
looks like it's up @lildziner! But yeah the code example isn't working. I'd recommend just linking to where folks can find the code if that's possible? Also are there any accessibility considerations or anything else that needs to be laid out on this page? https://design.va.gov/experimental-design/expandable-alert

PR #501, opened 2021-07-28
Code contributions to experimental designs

cvalarida:
Documentation for contributing code for experimental designs.

### DIFF
--- a/src/_experimental-design/contributing-code.md
+++ b/src/_experimental-design/contributing-code.md
@@ -2,18 +2,13 @@
 layout: default
 title: Contributing experimental design code
 order: 2
-draft: true
 anchors:
   - anchor: Writing experimental design code
-  - anchor: Code location
-  - anchor: README
-  - anchor: Breaking changes
-  - anchor: CODEOWNERS
-  - anchor: Test coverage
+  - anchor: Sharing experimental design code
   - anchor: Using experimental designs
 ---
 # Contributing experimental design code
-At the heart of the experimental design process is the shared code. This document explains the process for contributing code for experimental designs and the reasoning behind that process.
+This document explains the process for contributing code for experimental designs and the reasoning behind that process.
 
 If you haven't read it already, refer to the [experimental design page](/experimental-design) for more information about the full process.
 
@@ -26,6 +21,11 @@ Each experimental design should:
 Developing the experiment as if it were a standalone library will make the code more reusable and graduating the component or pattern into the official design system smoother.
 
 Each experimental design should [include a README](#readme) and be [owned by a team](#codeowners).
+
+## Sharing experimental design code
+Sharing code between applications is necessarily more involved than writing code for a single application. To avoid the overhead that sharing code introduces (reporting development status, managing breaking changes, deprecating the code etc.), it's recommended to develop experimental designs in the application directory and not import it from another application.
+
+If the time comes that another application needs to use the experiment, the rest of this section describes the process for how to share this code.
 
 ### Code location
 Each experimental design is located in its own directory in [`vets-website`](https://github.com/department-of-veterans-affairs/vets-website/) at `src/experimental/` unless otherwise noted in its documentation on this site.
--- a/src/_experimental-design/contributing-code.md
+++ b/src/_experimental-design/contributing-code.md
@@ -50,7 +50,7 @@ Once the code for an experimental design is stable, **breaking changes should no
 
 If you need to introduce breaking changes, **do not modify the existing code.** Instead, copy the contents of the directory to a sibling directory post-fixed with a version number.
 
-**Example:**
+**Example:**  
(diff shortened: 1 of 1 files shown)


src/_experimental-design/contributing-code.md
```diff
@@ -0,0 +1,73 @@
...
+If your team needs an experimental button that's larger than the standard button, you would create `src/experimental/large-button/index.jsx` as the entry file for your "library."
+
+### README
+Each experimental design should have a README that contains the following information:
+- Development status: `stable`, `unstable`, or `deprecated`
+     - The `unstable` status means the code is under active development and the public API may change without notice
+     - The `stable` status means the public API is finalized, but the code may still receive backward-compatible updates such as accessibility improvements and bug fixes
+     - The `deprecated` status means the code should no longer be used in applications
+          - This may be because of a breaking change (see [Breaking changes](#breaking-changes) below), official adoption into the design system, or research which indicates the experiment was unsuccessful
+          - When the status is `deprecated`, add a `Reason:` to indicate why (so engineers know where to look for the new version or not to use it at all) and make an announcement to alert other teams that may be using the experiment
```
cohnjesse, line 43: @cvalarida  I have worries about setting a precedent that we would have components lingering in the experimental design system that have already been moved into the main design system but simple have a developmental status of deprecated I would think a better way would be to simple have that code removed from the experimental design system, otherwise we are going to end up with folks not knowing what version of what to use.

cohnjesse (commented):
This document doesn't really tell folks much about how to contribute to the experimental design system and more introduces the idea of the development status. I feel like a lot more detail needs to be added to clarify when to add a component, what the process is for maintaining, etc...

src/_experimental-design/contributing-code.md
```diff
@@ -0,0 +1,73 @@
...
+Add your team's GitHub team name to the [CODEOWNERS file](https://github.com/department-of-veterans-affairs/vets-website/blob/master/.github/CODEOWNERS) to take ownership of the experiment's code. This will mean your team will be required reviewers on all changes to this code.
+
+### Test coverage
+As with all code, test coverage is critical. This is especially true with shared code. Aim for at least 90% unit test coverage before declaring an experiment to be `stable`.
+
+## Using experimental designs
+ The babel module resolver plugin has the `root` set to `"./src"`, so you can import your experimental design with the following:
+ 
+ ```js
+import LargeButton from 'experimental/large-button';
```
erikphansen, line 72: While this will work I think we should encourage people to use `~/experimental`

src/_experimental-design/contributing-code.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Code location
+Each experimental design is located in its own directory in [`vets-website`](https://github.com/department-of-veterans-affairs/vets-website/) at `src/experimental/` unless otherwise noted in its documentation on this site.
+
+**Example:**  
+If your team needs an experimental button that's larger than the standard button, you would create `src/experimental/large-button/index.jsx` as the entry file for your "library."
+
+### README
+Each experimental design should have a README that contains the following information:
+- Development status: `stable`, `unstable`, or `deprecated`
```
erikphansen, line 38: This section is the reason I'm uncomfortable with formalizing an experimental design system. IMO if your component is unstable, it shouldn't live here. Why encourage people to use it by putting it in a shared location? I would think code would only be here if it was relatively stable and another team needed it. And once multiple teams are using it from the experimental design system, updates should be made in ways so that it doesn't break underneath people. Which is another way of saying that components should only be marked at deprecated only if absolutely necessary.

erikphansen: Also, if following my own personal guideline of code only being here when it's relatively stable and used by multiple teams, I would hope the component would very quickly be promoted to the real component library.

src/_experimental-design/contributing-code.md
cvalarida: We talked about this in Slack, but I'll post here for posterity as well: Deprecated components are to be deleted when no applications use them any longer. The deprecated status is here to break up the work of changing applications which use the code. Otherwise, you'd have to delete the code _and_ change _all_ the applications in one PR. If your application is the only one using the code (then why is it shared code? :confused: ), then you can delete it outright.

I'll call out the end-of-life actions required, though. That's a good point. :+1:

src/_experimental-design/contributing-code.md
cvalarida: We talked about this in Slack, but I'll post here for posterity as well.

I 100% agree with you. This process is for sharing code. If a team is going to be the only consumer of an experimental design during the lifetime of the experiment, then it should live in the application code. (Still developed as a standalone component or pattern, of course, so it's easily portable. Because that's good, clean living. :stuck_out_tongue_winking_eye: ) When the next team comes along and wants to use it, that's when I'd recommend we go through all the added effort of moving it to a shareable location and be extra careful about the public API.

I'll call that out in the top of this document. Specifically that this whole process is for contributing code to be shared between applications and should only be shared when needed—because of the extra process involved.

PR #516, opened 2021-08-13
moved documentation page content

cohnjesse:
per [a Slack discussion](https://dsva.slack.com/archives/C01Q9PDPLE9/p1627324208041000) it was decided that we should move the experimental design system documentation for adding code to the "for developers" section. This PR does that.


cvalarida (commented):
Thanks for moving this over. :raised_hands: Only suggestion before merging is to adjust the rest of the heading levels. I see you made `Contributing experimental design code` an `h2`, which is great, but everything that was previously under it (`Writing experimental design code`, et al) are now sibling headings.

cohnjesse:
@cvalarida Updated the headings

cvalarida (approved):
:shipit:

Issue #511, opened 2021-08-10
Validation message variation


andaleliz:
## What

This is a variation of the error message we currently have on form inputs ([Storybook example](https://design.va.gov/storybook/?path=/docs/components-radiobuttons--error)).  The Authenticated Experience team developed this pattern to drive engagement with notification settings, and then give Veterans confirmation/error messages on their changes.

This pattern provides in-line feedback to users on forms that auto-save and differs from the current error pattern in that it:

- Includes an icon before the error message, so that we're not solely relying on color to communicate intent
- Includes a background color on the feedback message that ties to the border color that appears on the left side of the field set
- Includes success, informational and warning states, in addition to the current error state that we have.

We first tried using the current validation message pattern, but it doesn't work in a warning state (yellow text on white background), and relies solely on color.

## Why

> Explain why you think this should be added to the VA.gov design system.

This pattern gives designers an immediate way to provide inline feedback to Veterans related to input interactions.  The minimal padding inside the alert optimizes the pattern for mobile, especially when compared to existing alerts in the design system ([mock-up](https://preview.uxpin.com/ed1067b8f73e8f3501bc476bb03bb4b46a261a39#/pages/140958818/simulate/sitemap) of our page with the [background only alert](https://design.va.gov/storybook/?path=/docs/components-va-alert--background-only)).  The background color and icons give us means of visual communication to better associate the content with the intent of the feedback for sighted users.

The use case differs from our validation message pattern in a couple few ways:

- Current validation messages are driven only by front-end validation; this pattern will also update in response to API calls
- Current validation messages are used to tell the Veteran they've missed a required field, or they've entered data in the wrong format. This pattern can be used to communicate important information ahead of the Veteran interacting with an input, and display confirm/success messages when there is no button to trigger them otherwise.
- We are not using this pattern in a form with an explicit submit action. Making changes to these radio buttons will immediately trigger an API call.

> - What evidence do you have that it's needed by multiple services across VA?

We already use different patterns to communicate messages to users in forms (hint text, alerts, error validation). This is an alternative that could work for any form input where we need to communicate a message to user other than a user error message.  The different states allow for communication of different types of messaging in a more consistent way than what we do now.

It could also be used instead of the error message, since it has elements other than color to communicate intent.

> - What evidence do you have that it meets the needs of the users of those services?

[...]

PR #520, opened 2021-08-16
Create README.md

andaleliz:
Adding page to experimental design section regarding the input message pattern proposed in #511


andaleliz:
@sshein I'll add the images to this directory once this approved; I don't think I can upload them otherwise.

Issue #535, opened 2021-08-30
Checkbox error message placement is inconsistent 


andaleliz:
## What happened

The [guidance for the checkbox component ](https://design.va.gov/patterns/form-feedback#errors-on-checkboxes) advises inconsistent error message placement dependent on a single vs group of checkboxes.  

It's unclear if the component is programmed to reflect this behavior by default.

Based on [this Slack conversation](https://dsva.slack.com/archives/C0NGDDXME/p1630004593007700), there doesn't appear to be any reason for this inconsistency.

## What I expected to happen

- Error messages should appear consistently regardless of how many checkboxes are in a group - under the label, and above the first input (like we do for a checkbox group).
- Design system guidance recommends consistent placement of error messages.

When the error message appears, the focus should go to the error message so the user can arrow down to the first checkbox.

## Reproducing
- Device: Macbook
- Browser: Chrome

Steps to reproduce:

Review the the [guidance for the checkbox component](https://design.va.gov/patterns/form-feedback#errors-on-checkboxes).

andaleliz:
cc @GnatalieH for awareness

caw310:
Issue was discussed at the 9/10 DSC meeting. Follow up action items 
    * DST to compare existing Forms library component with DS component to make sure they’re the same. 
    *DST  Make announcement that checkbox component has been updated. Put out comms in Monday meeting and in slack that this is an error that has been corrected and to please check applications if you haven’t received this code.

PR #539, opened 2021-09-03
29551: Update error message placement for <Checkbox /> component

craigwheeler:
This PR moves the error message for the `<Checkbox/>` component in `src/_patterns/html/error-checkbox.html` to be located above the checkbox input rather than below it. Refer to this issue for details https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/535

### DIFF
--- a/src/_patterns/form-feedback.md
+++ b/src/_patterns/form-feedback.md
@@ -81,10 +81,6 @@ Error messages are generally the same. The error message is slightly different i
 
 ### Errors on checkboxes
 
-Unlike other form elements, error messages for checkboxes appear below
-
-#### Error message below checkbox
-
 <div class="site-showcase">
 {% include_relative html/error-checkbox.html %}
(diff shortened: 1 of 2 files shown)


craigwheeler:
change completed here https://github.com/department-of-veterans-affairs/component-library/pull/153 ready for docs to be updated

andaleliz:
Hey @craigwheeler thanks for taking care of this so quickly! It looks good to me. The only thing I'm questioning is the space between the error message and the new checkbox, but I think that would be a separate issue.

craigwheeler:
Hey @andaleliz 👋 ... yep I agree however that space is necessary since this component is used in places that might be dependent on the extra space so we determined it was best to leave it for now, you can reference the convo about this [here](https://github.com/department-of-veterans-affairs/component-library/pull/148#issuecomment-911943916)

PR #552, opened 2021-09-24
Use Web Component for fullwidth alert

bkjohnson:
## Description

Use the Web Component as an example now that we have one for `full-width`

## Screenshots

image


cvalarida (approved):
Awwwww yisssss. :shipit:

PR #555, opened 2021-09-29
Updated Table usage and accessibility guidelines

GnatalieH:
- Changed Usability to Usage
- Updated when to use/when not to use
- Added guideline about headers and accessibility


GnatalieH:
@bkjohnson I'd like to add more details about how to use proper table headers to support accessibility. I'm also interested in knowing if we actually use simple _and_ complex tables on va.gov, and I'd like to review the responsive table list guidelines, too. But I wanted to get this PR made before the close of the sprint.

PR #557, opened 2021-09-30
Create back-to-top-draft.md

GnatalieH:


src/_components/back-to-top-draft.md
```diff
@@ -0,0 +1,37 @@
...
+
+### When to use Back to top
+* In pages with static content that is more than two screens in length on large and medium screens, or more than four screens in length on small screens.
+
+### When _not_ to use Back to top
+* In pages with dynamic content.
+* In pages with static content that is short in length.
+
+### Placement
+* The Back to top component is anchored to the lower right edge of the main content container on desktop, and the lower right edge of the viewport on smaller screens.
```
bkjohnson, line 25: I think we should say something like:

> The Back to top component is anchored to the lower right edge of its _parent_ container.

So, we want to encourage people to place this component inside of the main content container so that it will be anchored to the appropriate right edge.

src/_components/back-to-top-draft.md
```diff
@@ -0,0 +1,37 @@
...
+sub_section: Back to top
+title: Back to top
+draft: true
+---
+
+# Back to top
+
+<p class="va-introtext">The Back to top component is a button shortcut that allows the user to quickly navigate to the top of a long page of content.</p>
+
+{% include storybook-preview.html height="300px" story="components-table--default" %}
```
bkjohnson, line 12: This can get updated to the appropriate story.

PR #568, opened 2021-10-29
updating Veteran entry

RLHecht:
@DanielleThierryUSDSVA @bethpottsVADEPO I took out the "awkward style" part of this entry and reworded the rest of the sentence a bit. Let me know what you think.

### DIFF
--- a/src/_content-style-guide/00_word-list.md
+++ b/src/_content-style-guide/00_word-list.md
@@ -508,7 +508,7 @@ vesicant agents: Use "vesicant (blistering) agents."
 
 Vet Center: Capitalize "Center."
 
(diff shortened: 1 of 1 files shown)


DanielleThierryUSDSVA (approved):
I like this, but that second sentence has a lot of clauses (as does the original). Even though our style guide isn't necessarily Veteran-facing content, I think we should try to follow our principles as much as possible here too.

How about this?

But whenever possible, use the second person point of view to speak directly with Veterans, rather than referring to them in the third person.

RLHecht:
Thanks @DanielleThierryUSDSVA. I changed this to your second suggestion.

PR #570, opened 2021-11-03
Adding info about zip codes

RLHecht:
@DanielleThierryUSDSVA @bethpottsVADEPO I've added the info about the extra four digits in a zip code to our address section. Let me know what you think.


GnatalieH:
Hi! Apologies if I missed this in the Slack thread, but do you think I should add any zip guidance to the [Address block](https://design.va.gov/components/address-block) documentation? I know there's not much content there, currently. But perhaps I could at least crosslink to the Content style guide entry on addresses. LMK what you suggest, thanks.

RLHecht:
@GnatalieH I think crosslinking to our style guide page would be great. I think it might also be a good idea to use a different example for the component that doesn't contain a zip with the extra four digits so it's not confusing to anyone that it's required.

PR #582, opened 2021-11-12
Update content-presentation.md

GnatalieH:
Starting a draft for adding Cards guidelines to Content presentation pattern.


GnatalieH:
Closing because Cards will be its own Pattern and not part of Content Presentation.

PR #591, opened 2021-12-03
Package updates

k80bowman:
Resolves all but one severe NPM vulnerabilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "extract-loader": "^3.0.0",
     "extract-text-webpack-plugin": "^4.0.0-beta.0",
     "file-loader": "^2.0.0",
-    "gulp": "^4.0.0",
+    "gulp": "^4.0.2",
     "gulp-clean": "^0.4.0",
     "gulp-rename": "^1.4.0",
     "gulp-string-replace": "^1.1.2",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "style-loader": "^0.23.1",
     "tar": ">=6.1.9",
     "url-loader": "^1.1.2",
-    "webpack": "^4.43.0",
+    "webpack": "^5.64.4",
     "webpack-cli": "^3.3.12",
     "webpack-dev-server": "^4.6.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,24 +33,24 @@
...
-    "gulp": "^4.0.0",
+    "gulp": "^4.0.2",
     "gulp-clean": "^0.4.0",
     "gulp-rename": "^1.4.0",
     "gulp-string-replace": "^1.1.2",
     "lodash": "^4.17.21",
     "mini-css-extract-plugin": "^0.4.4",
     "mixin-deep": ">=1.3.2",
-    "node-sass": "^4.14.1",
+    "node-sass": "^6.0.1",
```
bkjohnson, line 43: I didn't find `node-sass` used anywhere, so it _might_ be safe to remove.

image

package.json
```diff
@@ -33,24 +33,24 @@
...
     "raw-loader": "^0.5.1",
     "resolve-url-loader": "^3.0.0",
     "sass-loader": "^7.1.0",
     "set-value": ">=4.0.1",
     "style-loader": "^0.23.1",
     "tar": ">=6.1.9",
     "url-loader": "^1.1.2",
-    "webpack": "^4.43.0",
+    "webpack": "^5.64.4",
```
bkjohnson, line 51: I was still able to get the site to run with this major upgrade.

bkjohnson (approved):
I don't know if this is something we need to worry about - there were no major upgrades to `gulp`, but it isn't being found:

image

package.json
k80bowman: yeah, I was amazed that all I had to do was make one change to the webpack config and everything was fine.

k80bowman:
@bkjohnson that's interesting, my local build is able to find gulp...
image

bkjohnson:
Maybe it's an npm bug. This is what I got when I did an earlier install before the build where `gulp` wasn't found:

image

So, maybe things didn't install right the first time. I just re-installed and did a build and gulp is working fine!

k80bowman:
Oh interesting... look like I may need to update my node and npm versions.

humancompanion-usds:
I think this was superseded by the work in issue #679

PR #593, opened 2021-12-03
fix: Html encode query string

pdavies88:


src/assets/javascripts/components/search.js
```diff
@@ -66,7 +66,7 @@ var searchOnResultsPage = function() {
   if (typeof searchTerm != "undefined"){
     desktop_search_input.value = searchTerm;
     mobile_search_input.value = searchTerm;
-    document.getElementById('query-term').innerHTML = searchTerm;
+    document.getElementById('query-term').innerText = htmlEncode(searchTerm);
```
k80bowman, line 69: should we be encoding it before it gets put into the search input as well? (lines 67 & 68)

andresriveratoro: It looks like we may want to do that.
https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#rule-2-javascript-escape-before-inserting-untrusted-data-into-html-attribute-subcontext-within-the-execution-context

pdavies88:
Following guidance from: https://portswigger.net/web-security/cross-site-scripting/preventing

src/assets/javascripts/components/search.js
k80bowman: Could you save `htmlEncode(searchTerm)` to a variable instead of calling it three times?

pdavies88: Good call just updated with that

k80bowman (approved):
Looks great, thank you!

PR #599, opened 2021-12-09
Update for-designers.md

GnatalieH:
Removed download link for Pattern library in GitHub (it was out of date, and we want to encourage Sketch for Teams usage) and updated verbiage around accessing Sketch for Teams and the VA pattern library.


GnatalieH:
Put in up to date Sketch pattern library download link.

PR #617, opened 2022-01-06
Adding when not to bold bullet 

RLHecht:
We needed to add a new bullet bout not using bold when it's better to have an H3 or H4 because the label is acting as a subhead.


bethpottsVADEPO:
@RLHecht Can you make that very last sentence of the new bullet active voice?

RLHecht:
@bethpottsVADEPO I fixed that last sentence.

bethpottsVADEPO:
Thanks for changing that one line to active voice. Looks good1

PR #651, opened 2022-02-16
Copyedit of tabs document

RLHecht:
@GnatalieH let me know what you think of my changes.

### DIFF
--- a/src/_components/tabs.md
+++ b/src/_components/tabs.md
@@ -20,14 +20,14 @@ title: Tabs
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


GnatalieH:
> 

@RLHecht I think the changes look good, thanks! The only thing I'd leave the same is the **When to consider something else** header, because we use this same pattern (When to use X, When to consider something else) throughout our Component documentation, and I think it we should be consistent. Thanks again!

RLHecht:
@GnatalieH That totally makes sense regarding the header. I realized that earlier today when I was reading documentation about another component.

PR #664, opened 2022-03-11
Add relative links html

andresriveratoro:
I know this solution is not ideal as it creates another file to maintain, but I was unable to get it working otherwise.

Things I tried:
- Creating a new variable named contentGlobal and using `{% include {{ include.contentGlobal }} %}`
- Creating a new variable, capturing the value and passing it directly instead of using `include`

I kept running into `Invalid syntax for include tag. File contains invalid characters or sequences` build errors.


k80bowman (approved):
It's fine for now. I'm hoping to move this site out of Jekyll soonish, so hopefully we can get away from difficulties like these. Thank you for figuring this out!

Issue #678, opened 2022-03-29
Update ruby, packages, and gems used by documentation site


humancompanion-usds:
# Bug Report

- [x] I’ve searched for any related issues and avoided creating a duplicate issue.

## What happened

Ruby, Jekyll, and 23 other packages that allow us to build and deploy the documentation site are out of date. Apparently, updating Jekyll in the past has caused the site build to fail.

## What I expected to happen

I expected current versions of the packages and gems we depend on to build the site to be up to date. Updating Jekyll should not cause the site build to fail.

## Reproducing

I was not able to reproduce the failure using `npm build` or `npm start`, but still testing.

humancompanion-usds:
Pull request: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/679

humancompanion-usds:
Filed issue for the warnings, "Use calc() for calculations in formation/sass/base/_b-variables.scss":
https://github.com/department-of-veterans-affairs/va.gov-team/issues/39750

Filed issue for upgrading yarn in component-library:
https://github.com/department-of-veterans-affairs/component-library/issues/352

humancompanion-usds:
Filed issue for the sort ordering of the sidenav being different now:
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/691

PR #679, opened 2022-03-29
Jekyll and package updates

humancompanion-usds:
- Updates Jekyll and additional gems and packages. Also adds a ruby_version dotfile.
- Updates packages
- npm audit fix made magic

## Closes Ticket
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/678


humancompanion-usds:
I believe this will allow us to close pull request #591 (this PR goes to a newer version of node-sass and the same versions of webpack and gulp), #655 (url-parse appears to be removed) and #649 (there is a newer version in this PR for follow-redirects).

pdavies88:
> Things look fine, but when I run a local build Jekyll shows me some warnings:
> 
> image
> 
> Do we know why that is? I think this is new.

@bkjohnson so I think this is new too with us upgrading Jekyll. To me it reads that if you go to: http://127.0.0.1:4000/design/ for example we have a couple of html files here: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/tree/master/src/_design/html that could potentially conflict with the root html file that gets generated for us. That warning is for all the places that we have an html folder inside of a parent folder in src. I'm assuming doing these partials in this way is not preferred by the newest version of Jekyll.

humancompanion-usds:
Okay, moved us over successfully to Dart Sass with the last commit. There are 2 warnings from node_modules/@department-of-veterans-affairs/formation/sass/base/_b-variables.scss where we need to now use calc() which I'll file an issue for. 

The only other issue I can see is that we were depending on the odd filename sorting for collections where there is no ordering yml key in the Content section to get the Word List first and the Sign up last. That no longer works. We can file an issue for that as well. We'd need to either figure out how to sort on 2 properties (title and order) which I have not been successful with in the last hour. Alternatively, we can set an order on every markdown file or put the order in the config.yml in collections. I'm not keen on doing either of those in this issue. If the Content folks yell, they can yell at me and I'll explain the deal.

bkjohnson (commented):
With the most recent change I'm unable to build & run the site (I don't know if it's related to dart sass or not, I just happened to experience this after that commit).

Error when trying to serve jekyll

And if I run `bundle update --bundler`:

Error when trying to update bundler


Down in the log:

mkdir command not found

I get the same error related to `mkdir` when I run `gem install eventmachine -v '1.2.7' --source 'https://rubygems.org/'`.

bkjohnson:
Also, we may need to change some of this in order to deploy the site with CI:

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/ef2af5a757ad9bc2c021752e7745d85bdcdc8187/.circleci/config.yml#L10-L24

pdavies88:
So I did not have the same issues with the newest code and was able to get it to build locally using node 14.15.0 and ruby 2.7.5 
Once I pulled down the most recent updates I ran:
`npm install` then `bundle install` then `npm run start` and I was able to get the build running.

pdavies88:
> Also, we may need to change some of this in order to deploy the site with CI:
> 
> https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/ef2af5a757ad9bc2c021752e7745d85bdcdc8187/.circleci/config.yml#L10-L24

@bkjohnson I'm assuming we would need to update line 10 to 2.7.5

pdavies88:
Here are the new gems that got installed for me with the most recent update:
Screen Shot 2022-03-31 at 8 32 36 PM

For me when I ran `gem install jekyll` that's when eventmachine was installed for me:
Screen Shot 2022-03-31 at 8 37 30 PM

I did run `gem uninstall -aIx` to clear all gems from my system to retest when they get added.

Clearing everything out and then readding all the gems steps I ran were `npm install` | `gem install jekyll` | `bundle install` | `npm run start` and I was able to get the build running

pdavies88:
Installation instructions from Jekyll where they recommend `chruby`: https://jekyllrb.com/docs/installation/

bkjohnson:
After the OS install I can build this again :tada: 

Now I'm just seeing the deprecation warning for division which @humancompanion-usds pointed out:

image


Regarding the ordering, there's a PR (#452) which should let us do that. If we don't make [this change](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/commit/9bc5f4b5b84615544ffc4cb0f35bf8ab7a61574c#diff-22c19865adb60132c06705f0ac17025f0e82399ed9002fef4b476e4242ad219fR19), this looks like it would work without any changes to other markdown files or editing the `_config.yml`:

```diff
diff --git a/src/_content-style-guide/00_word-list.md b/src/_content-style-guide/00_word-list.md
index 577704f..5a1a40d 100644
--- a/src/_content-style-guide/00_word-list.md
+++ b/src/_content-style-guide/00_word-list.md
@@ -1,6 +1,7 @@
 ---
 layout: default
 title: Word list
+order: 1
 slug: word-list
 ---
 ```

image

**Review by k80bowman — APPROVED** [review]
I was able to run the site locally no problem. Thanks for the README updates!

File: Gemfile.lock [hunk 1]
```diff
@@ -4,76 +4,86 @@ GEM
...
+  jekyll (~> 4.2.2)
   jekyll-feed (~> 0.6)
   jekyll-last-modified-at
   minima (~> 2.0)
+  sass-embedded (~> 1.0)
   tzinfo-data
 
 BUNDLED WITH
-   1.16.6
+   2.1.4
```
bkjohnson, line 89: It looks like the `deploy-dev` job [worked just fine](https://app.circleci.com/pipelines/github/department-of-veterans-affairs/vets-design-system-documentation/1139/workflows/10b74e11-6c26-4d19-bcf2-94d9f2ddbbc6/jobs/2346), but it's still using version 1.16.6 for the bundler:

image

That's coming from here:

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/696f27c94e3afed2ffc294a241c5588ab5cc34ba/.circleci/config.yml#L24

It seems to be fine though :shrug:

humancompanion-usds:
> Regarding the ordering, there's a PR (#452) which should let us do that. If we don't make [this change](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/commit/9bc5f4b5b84615544ffc4cb0f35bf8ab7a61574c#diff-22c19865adb60132c06705f0ac17025f0e82399ed9002fef4b476e4242ad219fR19), this looks like it would work without any changes to other markdown files or editing the `_config.yml`:

@bkjohnson - When I remove the change I made and add order: 1 to Word List that does indeed but Word List first. However, in other sections I'm seeing what should be the last item in the list appear first. For example, Typography is first under Design, Telephone is the first component under Components, etc. Are you seeing the same thing? That's why I made the switch. I tried removing "last" from the ordering thinking perhaps there was a bug there but to no avail.

PR #682, opened 2022-03-30
Part2 of updating build fix README package.json and webpack config

pdavies88:
## Additional Work for Ticket
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/678

## Description of work performed
This PR updates the following that I was running into errors on when trying to run the base branch.
Updates:
- Clean up README.md to reflect change to ruby v2.7.5 and the use of node 14.15
- Update babel and its dependencies to the latest version to fix webpack errors
- Update webpack to fix errors with depreacted config and use css loader beyond leaving it as an orphaned import
- Remove deprecated code in entry.js from deprecated `formation-react`

Screenshots of Build:
Screen Shot 2022-03-30 at 2 40 12 PM
Screen Shot 2022-03-30 at 2 40 18 PM
Screen Shot 2022-03-30 at 2 40 26 PM
Screen Shot 2022-03-30 at 2 40 33 PM
Screen Shot 2022-03-30 at 2 40 38 PM
Screen Shot 2022-03-30 at 2 40 51 PM


pdavies88:
@bkjohnson general question do you think we need this file in its own folder: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/tree/master/webpack/components couldn't we just use it straight in `entry.js`? Also do we even need it (seems like boilerplate leftovers from the original build)?

pdavies88:
@humancompanion-usds:
The issues I ran into were the following after running the `npm run start` command (Node v14.15.0 and Ruby v.2.7.5):
- webpack-cli:
Screen Shot 2022-03-30 at 3 04 12 PM

- Switching this line to `options` which is the correct property https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/master/webpack.config.js#L17 lead to the following:
Screen Shot 2022-03-30 at 3 08 43 PM

- I then upgraded the babel dependencies to the latest to solve this and reran `npm install` and once again on `npm run start ran into this error: Screen Shot 2022-03-30 at 3 10 50 PM

- So then updated our `webpack.config.js` to be modeled after the one we are using on Component Library here: https://github.com/department-of-veterans-affairs/component-library/blob/master/packages/core/webpack.config.js
It also follows the webpack babel-loader guidelines: https://webpack.js.org/loaders/babel-loader/

pdavies88:
@humancompanion-usds @bkjohnson do you want me to bump our React version as part of this PR which would then get merged into base PR? Also which version do we want to bump it to: https://reactjs.org/versions/?

humancompanion-usds:
@pdavies88 - I leave that to you to decide. I'd imagine we'd want to match the same version in the component-library? Ideally, all parts of the Design System would be in-sync just for the sake of consistency and to avoid any possibility of there being some odd behavior due to a version mismatch.

pdavies88:
> @pdavies88 - I leave that to you to decide. I'd imagine we'd want to match the same version in the component-library? Ideally, all parts of the Design System would be in-sync just for the sake of consistency and to avoid any possibility of there being some odd behavior due to a version mismatch.

I would agree. Let me make the updates here and retest and then we can merge this branch into yours to do a final test to confirm everything is fully functional.

webpack/entry.js
```diff
@@ -1,78 +1,8 @@
...
- render() {
-   return (
-    <div>
-      <AdditionalInfo triggerText="Why is this required?">
-        <p>We need the Veteran’s Social Security number or tax identification number to process the application when it’s submitted online, but it’s not a requirement to apply for the program.</p>
-      </AdditionalInfo>
-    </div>
-   )
- }
+if (document.getElementById("reactComponentDemo")) {
```
bkjohnson, line 6: Do we need this file at all?

webpack.config.js
```diff
@@ -6,18 +6,27 @@ module.exports = {
   mode: "production",
   output: {
     path: path.resolve(__dirname, "src/vendor/javascripts/"),
-    filename: "bundle.js"
+    filename: "bundle.js",
```
bkjohnson, line 9: If this js bundle is just the `entry.js`, I'm not convinced that we need it. It that's true, we can probably also remove these to prevent some runtime errors:

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/4515e238a419a5287968a8d108838d7f13b544ee/src/_layouts/default.html#L116

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/4515e238a419a5287968a8d108838d7f13b544ee/src/_layouts/iframe.html#L10

bkjohnson (commented):
It's unclear to me what webpack is doing for us here and why we're making changes to webpacky things. If we can delete it I think that would be great.

webpack.config.js
pdavies88: @bkjohnson thats where my head went as well I'm not sure it's doing anything. Let me try to remove and see if we still get a functional build and I'll update you with the results

humancompanion-usds: The bundle.js it generates is referenced in the default layout file:
 <script type="text/javascript" src="{{ site.baseurl }}/vendor/javascripts/bundle.js" charset="utf-8"></script>

So just be sure to pull that out of there as well.

humancompanion-usds:
> It's unclear to me what webpack is doing for us here and why we're making changes to webpacky things. If we can delete it I think that would be great.

That's my fault. I was going about this by upgrading first and then had planned to tackle revising the build process to remove any cruft rather than doing them both at the same time. I thought getting to updated versions would make it easier to assess how to improve the build and what was safe to remove. My end goal is to crib some of the [nicer elements of what USWDS has](https://github.com/uswds/uswds-site/blob/main/package.json) (browserify, gulp-dart-scss, etc.) for their build process.

PR #701, opened 2022-04-15
Update tables.md

GnatalieH:
@bkjohnson I added a sentence to one bullet point under the (non-responsive) Table's Accessibility considerations to talk about the title. Does this look correct?


bkjohnson (approved):
Yes, this looks correct :+1:

bkjohnson (commented):
Actually, we may want to update that first sentence. We can say that the `table-title` prop will put its value into a `<caption>` tag.

PR #721, opened 2022-04-21
update component-docs logic

k80bowman:
Updates the component-docs logic so that when a component does not have any props or events, the code usage section is not added to the page.

### DIFF
--- a/src/_includes/component-docs.html
+++ b/src/_includes/component-docs.html
@@ -1,8 +1,8 @@
 {% for component in site.data.component-docs.components %} {% if component.tag
   == include.component_name %}
-{% unless compoenent.props == empty %} {% unless component.events == empty %}
+{% if component.props != empty or component.events != empty %}
 <h2 id="code-usage">Code usage</h2>
-{% unless component.props == empty %}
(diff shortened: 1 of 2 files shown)


src/_includes/component-docs.html
```diff
@@ -1,4 +1,8 @@
+{% for component in site.data.component-docs.components %} {% if component.tag
+  == include.component_name %}
+{% unless compoenent.props == empty %} {% unless component.events == empty %}
```
andresriveratoro, line 3: This looks like a typo for `component.props`

humancompanion-usds (approved):

src/_includes/component-docs.html
andresriveratoro, line 3: If I'm understanding this correctly, nothing will be shown if a component has no events? `va-table` has properties but no events.

k80bowman: I spotted that when trying to fix something else. I've updated the logic and made sure that `va-table` works now. Thank you!

pdavies88 (approved):

andresriveratoro (approved):

Issue #732, opened 2022-04-22
Add layout for component category pages on design.va.gov


k80bowman:
## Description
In the mural board with the new IA for design.va.gov, several of our components are combined into a navigation category. Those components are still on their own pages, though, so we will need a category page to appear at the top level and link to the individual component pages.

We would like to develop this as a layout that can be used for other groups of components  as well, so it should not be tied only to the navigation category.

## Details
IA Mural board

Navigation section architecture
![image.png](https://images.zenhubusercontent.com/5952ac1861b4ca3b0a727ca8/f0fd6c14-e7ab-43a3-8d8d-fb9f7d32b930)

## Tasks
- [ ] Check with Matt and Natalie about the design for this page and what needs to be included
- [ ] Create layout file
- [ ] Add layout to category page for the navigation elements

## Acceptance Criteria
- [ ] The layout file for component category pages exists
- [ ] The Navigation category page is using the category page layout

humancompanion-usds:
The Content team mocked up what this might look like, with a few caveats:

![Screen Shot 2022-04-22 at 4.31.03 PM.png](https://images.zenhubusercontent.com/623c98bb27734da1243c0bb5/91c7606e-363b-494e-bfcc-c2aae5dd8d68)

* We would expand the sidenav to reveal the sub-pages
* The title here would just be "Navigation" (we're already in the Components section so no need to repeat that)
* Some of these landing pages may include general guidance of their own. For example, "Form" would likely contain something like this: https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/teams/vsa/design/getting-started-with-va.gov-forms.md That's quite lengthy so we have to think about how much to include but the point is the content won't necessarily just be one paragraph. So my thought is to do another "intro-text" for the leading paragraph and then the navigation to the sub pages and then the full content.

humancompanion-usds:
I need to build a few more of these before we land on a specific template. These may need to remain super flexible. Let's revisit once we have the proper sidenav in place.

humancompanion-usds:
I've created a Category template: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/master/src/_layouts/category.html

As I suspected, at the moment we need these to be flexible because we have different uses. Putting this back on the backlog for now.

PR #734, opened 2022-04-25
sub pages 33924

humancompanion-usds:
- Allows the sidenv to have sub-pages. Breaks up Patterns - Forms into sub pages. [Amends #33924]
- Properly moves Design to Foundation. Moves Layout and Utilities to be sub-sections of Foundation. Makes Templates a sub-section of Patterns. [Amends #33924]


k80bowman:
This looks fine. The only issue I'm seeing is that the only way a user can close one of the subsections in the side nav is to open a different one. Is that something you want to try to fix now or as part of a different PR?

humancompanion-usds:
Yeah, it's a hack for sure. The real sidenav on va.gov is an accordion and thus the current index pages would become Overview pages inside their section so that the accordion face is just doing the job of open and closing the sub-nav rather than also acting as a target for that Overview page. I expect that when we do issue #732 and the sidenav (which I can't find right now) that we'll correct these problems. If that can get done this week during the testing that would be fine because I think the behavior change is minor enough that it would not impact the navigation tasks I have queued up in the testing. But it's also not a requirement.

k80bowman (approved):

PR #740, opened 2022-04-26
Fixes search. [Amends #732]

humancompanion-usds:
There is an odd line at the end of the search.json that I removed thinking it was innocuous. Wrong!


k80bowman (approved):

src/_data/component-docs.json
```diff
@@ -1302,190 +1302,6 @@
       "dependencies": [],
       "dependencyGraph": {}
     },
-    {
-      "filePath": "./src/components/va-number-input/va-number-input.tsx",
```
bkjohnson, line 1306: @humancompanion-usds is there a reason why we're committing changes to this instead of having it in the gitignore? For example, we probably don't want to remove component docs for the number input.

PR #743, opened 2022-04-26
fix utilities links and replace iFrames

k80bowman:
Fixes links going to foundation/utilities and replaces iFrames in utilities section with partials.


humancompanion-usds (approved):
Thank you!

PR #758, opened 2022-04-27
chore: fix links for In This Section

andresriveratoro:
This PR removes the extra slash in URLs in `In This Section`.

Screen Shot 2022-04-27 at 09 20 25
Screen Shot 2022-04-27 at 09 20 18


pdavies88 (approved):
That dastardly slash

PR #765, opened 2022-04-27
Fit content inside iframe

andresriveratoro:
## Screenshots
Screen Shot 2022-04-27 at 18 33 23
Screen Shot 2022-04-27 at 18 33 08

## Alternative Solution
We might be able to set the iframe height dynamically to fit its contents, but I haven't been able to test it locally because of cross-origin errors (Storybook runs on a different port locally). I think this would be considered "same-origin" on production? @bkjohnson @pdavies88 @k80bowman 
Screen Shot 2022-04-27 at 18 36 12

```diff
diff --git a/src/_includes/storybook-preview.html b/src/_includes/storybook-preview.html
index 1d19f9a..e862345 100644
--- a/src/_includes/storybook-preview.html
+++ b/src/_includes/storybook-preview.html
@@ -3,10 +3,18 @@
   style="max-width: {{ include.width | default: '100%' }}"
 >
   <iframe
+    id="{{ include.story }}"
     width="100%"
     height="{{ include.height }}"
     src="{{ site.storybook_path }}/iframe.html?id={{ include.story }}&viewMode=story"
   ></iframe>
 </div>
 
+<script>
+  window["{{ include.story }}"] = document.getElementById("{{ include.story }}");
+  window["{{ include.story }}"].onload = function() {
+    window["{{ include.story }}"].style.height = window["{{ include.story }}"].contentWindow.document.body.scrollHeight + "px";
+  }
+</script>
+
 [{% if include.link_text %}View {{ include.link_text}}{% else %}See this{% endif %} in Storybook]({{ site.storybook_path }}/?path=/docs/{{ include.story }})
```

### DIFF
--- a/src/_components/alert.md
+++ b/src/_components/alert.md
@@ -47,7 +47,7 @@ Used to prompt a user to sign in, create an account, or launch an online tool to
 
 Any style of alert box can be modified to be a background-color-only alert. Use background alerts for immediate feedback, such as in single-page applications or Ajax forms. They shouldn’t be used for static alert messaging.
 
-{% include storybook-preview.html story="components-va-alert--background-only" height="80px" %}
+{% include storybook-preview.html story="components-va-alert--background-only" height="352px" %}
 
 - Some users might not be able to distinguish differences in the background color or see the color at all. Don’t rely on color alone to convey context. 
 - Messaging should be direct and concise. Aim for one or two lines.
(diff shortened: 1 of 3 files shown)


humancompanion-usds:
@andresriveratoro - It's currently "not working" in that the fixed height hides the contents so I would say give your alternative a try. If it works, great! Then you could put that script block into a more permanent locale and update other iframes as needed. If it doesn't, then no harm done. Just move onto fixing the hard-code heights to be tall enough in Alert.

PR #766, opened 2022-04-28
Fix broken links in About

andresriveratoro:
This PR fixes broken links in the About section by removing trailing slashes and updating the URL to point to the new sections.


andresriveratoro:
There are more broken links in the `What's new?` section. I stopped at `April 5, 2021` because we should probably prioritize more important pages for now. The `form template documentation` link under `April 5, 2021` is broken, but I left it untouched because I'm not sure what the new link should be.

pdavies88:
@k80bowman in what's new do you think we should keep things that are over a year old? Mainly I'm thinking they aren't really new anymore.

pdavies88 (approved):

humancompanion-usds:
Let's shorten up the "What's new?" to just contain entries from the last 6 months.

PR #770, opened 2022-04-28
Adds role of header to dts, fixes up in this section, styles on this …

### LINKED ISSUE #768: "On this page" definition title doesn't have heading role
# Bug Report

- [x] I’ve searched for any related issues and avoided creating a duplicate issue.

## What happened

design.va.gov has a copy-pasted "On this page" component modeled after `<va-on-this-page>`. Additionally, there is a "Resources" section which is modeled after the "On this page" definition list:

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/4f33e589d5b5785282d762c081767345eeb01d04/src/_includes/component-resource-links.html#L13-L16

During a research session this section was not findable by a screen reader due to a missing heading. The "Possible fixes" section of [the original ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/252) had a `role="heading"` on the `<dt>` which wasn't implemented with the other definition list changes. Adding this attribute should allow the element to be appropriately flagged in the accessibility tree.

## What I expected to happen

I expected "On this page" and "Resources" to be findable by a screen reader when scanning for headings.

humancompanion-usds:
…page to match the component. [Fixes #768]

### DIFF
--- a/src/_includes/_site-in-this-section.html
+++ b/src/_includes/_site-in-this-section.html
@@ -1,6 +1,6 @@
 {% if page.sub-pages %}
 <nav aria-labelledby="in-this-section" class="site-in-this-section">
-  <dt id="in-this-section" role="header" class="site-in-this-section__title">
+  <dt id="in-this-section" role="header" aria-level="2" class="site-in-this-section__title">
     In this section
   </dt>
   <dl class="site-in-this-section__list">
--- a/src/_includes/_site-in-this-section.html
+++ b/src/_includes/_site-in-this-section.html
@@ -1,7 +1,9 @@
 {% if page.sub-pages %}
 <nav aria-labelledby="in-this-section" class="site-in-this-section">
-  <h2 id="on-this-page" class="site-in-this-section__title">In this section</h2>
-  <ul class="site-in-this-section__list">
+  <dt id="in-this-section" role="header" class="site-in-this-section__title">
+    In this section
+  </dt>
+  <dl class="site-in-this-section__list">
     {% for item in page.sub-pages%}
     <dd role="definition" class="site-in-this-section__definition">
       <a
@@ -12,6 +14,6 @@
       </a>
     </dd>
     {% endfor %}
-  </ul>
+  </dl>
 </nav>
 {% endif %}
--- a/src/_includes/component-resource-links.html
+++ b/src/_includes/component-resource-links.html
@@ -11,7 +11,7 @@
 
 <nav aria-labelledby="component-resources" class="site-component-resources-links">
   <dl class="site-component-resources-links__list">
-    <dt id="component-resources" role="header" class="site-component-resources-links__title">
+    <dt id="component-resources" role="heading" aria-level="2" class="site-component-resources-links__title">
       Resources
     </dt>
     <dd role="definition" class="site-component-resources-links__definition">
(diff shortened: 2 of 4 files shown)


src/_includes/_site-in-this-section.html
```diff
@@ -1,7 +1,9 @@
 {% if page.sub-pages %}
 <nav aria-labelledby="in-this-section" class="site-in-this-section">
-  <h2 id="on-this-page" class="site-in-this-section__title">In this section</h2>
-  <ul class="site-in-this-section__list">
+  <dt id="in-this-section" role="header" class="site-in-this-section__title">
```
bkjohnson, line 3:
```suggestion
  <dt id="in-this-section" role="heading" class="site-in-this-section__title">
```

src/_includes/component-resource-links.html
```diff
@@ -11,7 +11,7 @@
 
 <nav aria-labelledby="component-resources" class="site-component-resources-links">
   <dl class="site-component-resources-links__list">
-    <dt id="component-resources" class="site-component-resources-links__title">
+    <dt id="component-resources" role="header" class="site-component-resources-links__title">
```
bkjohnson, line 14:
```suggestion
    <dt id="component-resources" role="heading" class="site-component-resources-links__title">
```

bkjohnson (commented):
I think `heading` is the [role we want](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques).

humancompanion-usds:
@bkjohnson - Got it. Thanks. I should have checked instead of trusting the original issue that was filed. However, now that I've checked Mozilla essentially says not to do this at all: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role

I'm guessing we use a definition list and no header tag (h2-h6) because we don't want to make the On this page a h2 or h3 and disrupt the headings on the rest of the page. I also see our On this page web component doesn't set a role nor an aria-level either. Seems like we should based on the Mozilla page.

bkjohnson:
> Mozilla essentially says not to do this at all

From the [linked docs under "Best practices"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#best_practices):

> The best way to use this role is to not use it at all, and instead use the native heading tags [`<h1>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) through [`<h6>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements)

We [originally had an `<h2>On this page</h2>`](https://github.com/department-of-veterans-affairs/component-library/pull/59/files#diff-d34ae048e24b5aefb7413f8920a644a39ca9163b3b487ca38c1de1f21c54442dR20) since the component is designed to be the first "top-level" navigation element after the `<h1>`.

---

>  I also see our On this page web component doesn't set a role nor an aria-level either. Seems like we should based on the Mozilla page.

Correct, but I think the role part was a mistake. The [original ticket that led to switching from an `<h2>` to a `<dl>`](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/252) had that included in the "Possible fixes":

image

But the [PR for the issue](https://github.com/department-of-veterans-affairs/component-library/pull/169/files) did not include that change. It's still unclear for me why the switch to a `<dl>` was made - I wasn't involved in that PR.

Adding `aria-level` sounds like a good enhancement if we stick with `role="heading"`.

PR #793, opened 2022-05-03
Use GitHub action for doc site deployment

bkjohnson:
## Description

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/33568

This creates a GitHub action for the design.va.gov deployment which is currently handled with CircleCI

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,54 @@
+
+name: Deploy site
+on: push
+
+jobs:
+  deploy:
+    strategy:
+      matrix:
+        bucket: [dev-design.va.gov, staging-design.va.gov]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: master
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14.x
+          cache: 'yarn'
+      - run: npm install
+      - run: gem install bundler:2.1.4
+      - run: bundle install
+      - run: npm build
+      - run: bundle exec jekyll build --config _config.yml,jekyll-configs/dev.yml
+      - name: Remove .html extension on non-index files
+        run: |
+          find _site/ -type f ! -iname 'index.html' -iname '*.html' \
+          -print0 | while read -d $'\0' f; do mv "$f" "${f%.html}"; done
+      - name: Sync extensionless html files with correct type
+        run: |
+          aws s3 sync _site s3://<< parameters.bucket >> \
+          --acl public-read \
+          --delete \
+          --exclude "storybook/*" \
+          --exclude "*.*" \
+          --content-type "text/html"
+       - name: Sync remaining files
+         run: |
+          aws s3 sync _site s3://<< parameters.bucket >> \
+          --acl public-read \
+          --delete \
+          --exclude "*" \
+          --include "*.*" \
+          --exclude "storybook/*"
+
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: "us-gov-west-1"         
+      - name: S3 Sync Storybook build
+        run: |
+          aws s3 sync ./packages/storybook/storybook-static s3://${{ matrix.bucket }}/storybook --acl public-read --follow-symlinks --delete
+ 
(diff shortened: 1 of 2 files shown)


src/_components/index.md
```diff
@@ -4,7 +4,9 @@ title: Components
 index: true
 ---
 
-# Components
+# Components TESTING
+
+Some stuff here - deployed to dev?
```
bkjohnson, line 9: Testing that changes actually get deployed:

Changes appearing on deployed page

.circleci/config.yml
```diff
@@ -1,94 +0,0 @@
...
-            - Platform
-  deploy-prod:
-    jobs:
-      - build-deploy:
-          # Set the bucket parameter to be the prod bucket
-          bucket: "design.va.gov"
-          envConfig: "jekyll-configs/prod.yml"
-          filters:
-            branches:
-              only: master
```
bkjohnson, line 91: Previously we had this same config for the dev, staging, & prod jobs, so we were only running them when something got pushed to `master`.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,51 @@
+name: Deploy site
+on:
+  push:
+    branches:
+      - master
```
bkjohnson, line 5: This should keep that same functionality. [GitHub action documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags) on including branches.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  deploy:
+    strategy:
+      matrix:
+        environment: [{ bucket: dev-design.va.gov, config: dev.yml }, { bucket: staging-design.va.gov, config: staging.yml }, { bucket: design.va.gov, config: prod.yml }]
```
bkjohnson, line 11: The map values inside the `environment` array represent the `bucket` and `envConfig` parameters from the CircleCI job.

The matrix strategy will run one job for each of the environments. I plan on writing a ticket for future work which may improve this.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js 14.x
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14.x
+        cache: 'npm'
+    - uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.7.5 # Not needed with a .ruby-version file
```
bkjohnson, line 22: We have a `.ruby_version` file, and I don't know enough about our ruby setup to know how that's used and if it can be safely renamed. So we need to leave this in for now.

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/06429e0f41fb9929bff6e7240319f9769e23a833/.ruby_version#L1

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,51 @@
...
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - run: npm install
+    - run: npm run-script build
```
bkjohnson, line 25: The CircleCI config was using yarn for this. I switched to npm thinking that there may be a caching benefit since npm is what the repo uses and we have a `package-lock.json` file, not a `yarn.lock`.

.circleci/config.yml
```diff
@@ -1,94 +0,0 @@
...
-            --content-type "text/html"
-      - run:
-          name: Sync remaining files
-          command: |
-            aws s3 sync _site s3://<< parameters.bucket >> \
-            --acl public-read \
-            --delete \
-            --exclude "*" \
-            --include "*.*" \
-            --exclude "storybook/*"
```
bkjohnson, line 55: Commenting on this chunk to pull it into the comments.

.circleci/config.yml
```diff
@@ -1,94 +0,0 @@
-# Use the latest 2.1 version of CircleCI pipeline process engine. See: https://circleci.com/docs/2.0/configuration-reference
-version: 2.1
-# Use a package of configuration called an orb.
-orbs:
-  aws-cli: circleci/aws-cli@1.3.1
```
bkjohnson, line 5: This combined with the `Platform` context (contains aws credentials) allowed the `aws` commands to run.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,56 @@
...
+        --exclude "*.*" \
+        --content-type "text/html"
+    - name: Sync remaining files
+      run: |
+        aws s3 sync _site s3://${{ matrix.environment.bucket }} \
+        --acl public-read \
+        --delete \
+        --exclude "*" \
+        --include "*.*" \
+        --exclude "storybook/*"
```
bkjohnson, line 56: This is a copy of what the CircleCI config was doing (highlighed above for comparison) with the addition of the `aws-actions/configure-aws-credentials` step right before we run the `aws` commands.

bkjohnson (commented):
Pointing out the similarities & differences between the two configs.

Issue #775, opened 2022-04-28
Indicate TTY numbers in va-telephone


bkjohnson:
# Description

Some phone numbers are only meant to be used with a TTY. The `<va-telephone>` component should have a way of indicating this.

## Context

There was a [slack thread](https://dsva.slack.com/archives/C8E985R32/p1651155317299029) about how to format an `<a>` tag for TTY use. If `<va-telephone>` had this functionality we could recommend that the component be used.

## Questions

- Should the "TTY" text be included in the text inside of the `<a>` tag as well

## Acceptance Criteria

- [ ] `<va-telephone>` compnent indicates that the number is a TTY number in the `aria-label`

k80bowman:
Let's make sure that the issues reported in department-of-veterans-affairs/va.gov-team#40492 and department-of-veterans-affairs/va.gov-team#43969 are addressed as part of the work for this ticket.

k80bowman:
[Guidance on telephone numbers from design.va.gov](https://design.va.gov/content-style-guide/dates-and-numbers#phone-numbers)

PR #800, opened 2022-05-03
Updating TTY entry

RLHecht:
I updated the entry for TTY numbers so that TTY is also included in the linkable text.


k80bowman (approved):

bkjohnson (commented):
Related ticket: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/775

Also, should we have the `dates-and-numbers` page tell people to use `<va-telephone>` instead, with an exception for TTY numbers?

humancompanion-usds:
> Related ticket: #775
> 
> Also, should we have the `dates-and-numbers` page tell people to use `<va-telephone>` instead, with an exception for TTY numbers?

Yes please!

RLHecht:
@bkjohnson I'm not sure I understand what this is. We don't use va-telephone as a component in Drupal. Can you show me an example of what this is? Not sure if this is something for the style guide if it's not used everywhere.

bkjohnson:
It's a component from the VA Design System and it's used in several places in application code. Here's [one example](https://github.com/department-of-veterans-affairs/vets-website/blob/9243d96ed6511c56afe775c165a00e316978e35d/src/applications/disability-benefits/996/content/GetFormHelp.jsx#L11-L16) which can be found on the [start page of the "Request a Higher-Level Review" form](https://staging.va.gov/decision-reviews/higher-level-review/request-higher-level-review-form-20-0996/start):

Highlighted `<va-telephone>` in "Need Help" footer section

This specific component isn't used in the liquid templates _yet_, but other components like it are. I'm not familiar enough with the Drupal preview server to know if it's setting up our component library, but if that component is used in Drupal it will work on va.gov. If the Drupal preview server doesn't show these components properly and that's considered a blocker, then I'm happy to answer questions about how to get the components to work.

RLHecht:
@bkjohnson We don't usually put design system components into the content part of the style guide since this is for content authors across VA.gov, who aren't necessarily using application code. Though I think the code you showed in the example above would need to be updated so that the TTY part is linkable. Is that something you or someone else can update in the design system? 

I think this PR can be merged as is since the content update has been made. FYSA @bethpottsVADEPO

bkjohnson:
Design system components can be used outside of application code.

To make the "TTY" part linkable the DST would complete work for the ticket I linked earlier, and then the app team would update their code accordingly.

bethpottsVADEPO:
Let's merge this PR. I'd like to have a separate conversation about whether we should start using <va-telephone> on our static pages that live in Drupal. Thanks!

PR #809, opened 2022-05-05
Internal templates

k80bowman:
Adds a basic issue template and a set of templates for Design System component work. All are labeled for internal DST use only.


.github/ISSUE_TEMPLATE/component_development.md
```diff
@@ -0,0 +1,37 @@
...
+- Review [Design System Engineering Best Practices document](https://vfs.atlassian.net/wiki/spaces/DST/pages/2176516116/Design+System+Engineering+Best+Practices)
+
+## Tasks
+- [ ] Create web component and add to Storyboook
+- [ ] Write any necessary tests
+- [ ] Add Chromatic link to #[add accessibility ticket number] and request review from an accessibility specialist
+- [ ] Ping designer for design review
+- [ ] Once accessibility and design reviews are complete, alert Carol that component is ready for staging review
+- [ ] Address any staging review comments
+- [ ] Set maturity scale to "Use with Caution: Available"
```
andresriveratoro, line 31: Does this apply to web components that are replacing their React counterparts as well?

k80bowman: Ah good catch, thank you! I changed that in another ticket and forgot to update it here.

pdavies88 (approved):

.github/ISSUE_TEMPLATE/component_announcement.md
```diff
@@ -0,0 +1,41 @@
...
+Dependent on the completion of development, the creation of Sketch artifacts and component page, and documentation for the component on design.va.gov.
+
+In the Slack announcement, please add links to the component page on design.va.gov and to the Design System Maturity Scale page on design.va.gov. Post the announcement in the following Slack channels:
+- #vfs-all-teams
+- #vfs-engineers
+- #vfs-frontend
+- #design
+
+Sample announcement:
+> :loudspeaker: The Design System has just published a new version of component-library which includes the new [component name] component, as well as made it available in vets-website. This component is rated Use with Caution: Available on the Design System Maturity Scale and the Design System Team will be monitoring its usage over the next few months.
```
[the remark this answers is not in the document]
humancompanion-usds: I just want to hammer home here that we are releasing a component across design.va.gov, Storybook, and Sketch at the same time to quell any complaints that those are out-of-sync.

humancompanion-usds (approved):
This is a really good start. We'll need some of these to be duplicated for a Pattern at some point. Also, we need a checklist for testing a Component before it gets announced across design.va.gov, Storybook, and Sketch. In some ways it is just a master checklist that tracks progress across these templates. But there also should be a final review step before we go to Staging in the CollabCycle. I'd like a chance to review things before we go ask to publish them to make sure they meet our team standards (which should exceed, as much as possible, the CollabCycle standards).

k80bowman:
@humancompanion-usds I made the updates you requested. I really like your idea about a checklist for tracking progress of a component through the templates. We could probably set something up in Confluence, what about something like [this](https://vfs.atlassian.net/wiki/spaces/~752492761/pages/2188083248/Component+Progress+Tracker+-+DRAFT)?

As far as a final review step before Staging Review, I agree that's a good idea. Right now, I have Staging Review as a task under the Development ticket because we typically don't merge component code until Staging Review is done. But, if we want to go through a more thorough review before scheduling Staging Review, maybe we could make use of the Maturity Scale. We could label it "Use with Caution: Candidate" until Staging Review is complete and then update it to "Use: Deployed". Then we could have an "Review" ticket that includes a checklist of things we want to make sure have happened prior to Staging Review. What do you think?

humancompanion-usds:
@k80bowman - Yes, that would be a good use of the maturity model. We can signal something is in-work with "Use with Caution: Candidate", especially if it's work that's come through the DSC and/or experimental process, that would be the right thing to do. 

That Tracker you linked to is fine for an overall tracker for our team. We were discussing this morning a more detailed version we might put in the system itself which would track things like if and when the component has been checked for a11y, is up-to-date in Storybook and Sketch (and if not detailed what is aberrant), has all variations represented, etc. I think we'll try and co-create this at some point.

k80bowman:
@humancompanion-usds Great! I have updated several of the tickets accordingly and added a final review ticket. Let me know what you think.

Just to document, this is what has been in my mind as the workflow for the component development process. It's based on some team discussions we had a few months ago, along with the feedback I've gotten from you, and I'd love to hear any additional thoughts you have.

image

I love the idea of having a more detailed component checklist to add to the DS documentation, we should definitely work on that. Maybe that could be part of Phase II of our documentation improvement work.

PR #839, opened 2022-05-10
Plain language page

humancompanion-usds:
- Creates a layout for the content-style-guide section
- Makes plain language and child of content principles
- Figured out how to pass markdown to the Like this, Not this partial


bethpottsVADEPO (approved):
Looks good, thanks so much!

PR #841, opened 2022-05-10
Changes the source order of the maturity icons. 

### LINKED ISSUE #777: Improve accessibility of component maturity sidenav indicator on design.va.gov
## Description
The component maturity indicator in the sidenav is creating issues for voiceover users. According to Angela Fowler (in a user research session) it would be better if the indicator was semantically after the component name in the sidenav. However, for design purposes, we want it to visually appear before the component name.

## Details
Brooks suggested using flex-order to visually put it in front of the link text but semantically after the link text.

## Acceptance Criteria
- [ ] The component maturity indicator is semantically after the component name but appears visually ahead of it
- [ ] The maturity indicator is a different shape for use, use with caution, and don't use

humancompanion-usds:
Changes the shapes of some of the maturity icons. Carries those new shapes into the maturity level pages. Fixes a minor incorrect reference on the link page (could not help myself). [Fixes #777]

### DIFF
--- a/src/_about/maturity-scale.md
+++ b/src/_about/maturity-scale.md
@@ -16,19 +16,32 @@ anchors:
 
 The maturity scale is divided into 3 major categories:
 
-* <span class="usa-label site-component-status site-component-status--use">Use</span>
-* <span class="usa-label site-component-status site-component-status--use-with-caution">Use with caution</span>
-* <span class="usa-label site-component-status site-component-status--dont-use">Don't use</span>
+  
+{% assign scale = site.data.maturity-scale.scale %}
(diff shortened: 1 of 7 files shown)


pdavies88 (approved):

humancompanion-usds:
@bkjohnson - Yes, the older versions of Font Awesome did not do a good job of normalizing the icons so it has this problem. In fact, there is no triangle, and really no other discernible shape to differentiate from circle or square. So I ultimately decided that it was acceptable because I just wanted to change the shape and one can make out the overall shape. If we don't go to USWDS v3 icons in future then we should at least update Font Awesome to the latest and greatest free version.

PR #851, opened 2022-05-11
745 design system sidenav

pdavies88:
## Description
Closes Ticket https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/745

### Mobile:
Screen Shot 2022-05-11 at 11 55 41 AM
Screen Shot 2022-05-11 at 11 55 59 AM

### Desktop:
Screen Shot 2022-05-11 at 11 56 43 AM
Screen Shot 2022-05-11 at 11 57 15 AM
Screen Shot 2022-05-11 at 11 57 02 AM

### Considerations:
To see how secondary navigation works on va.gov click [here](https://www.va.gov/health-care/eligibility/) and tertiary click [here](https://www.va.gov/health-care/eligibility/active-duty/). I personally don't like losing the whole sidenav when working on a tertiary page which is why I went forward with my approach.

Additionally on **Mobile** for design.va.gov we create a tertiary level with our code. I could build that in to match desktop if we would likem but right now it just flattens the pages at a secondary level ie: Form and Checkbox live on the same level even though on desktop Checkbox is a child of Form.
Screen Shot 2022-05-11 at 11 56 17 AM

## Acceptance Criteria
- [x] Adapt the Design System sidenav component markup in the design.va.gov sidenav component file
- [x] Test to make sure nested nav items functions correctly


src/_foundation/utilities/text-color.md
```diff
@@ -1,11 +1,11 @@
 ---
 layout: default
-permalink: /foundation/utilities/color
+permalink: /foundation/utilities/text-color
```
pdavies88, line 3: Changing the name to better represent what the file does and I was also getting a conflict with active states when trying to differentiate between `background-color` and `color` since both strings would match in the contains operator when parsing the url

pdavies88:
@humancompanion-usds @k80bowman feel free to let me know if you have feedback on the `Considerations` section I put into the ticket

humancompanion-usds:
@pdavies88 - I managed to reconcile the conflicts and reapply the a11y fix using flex and order to place the icon after the text in source order but visually put the icon in front of the text. I also managed to add icons for sub-list items but that appears to only be working for sub-items in Components and not in Patterns and I have no clue as to why. Can you look into that? Otherwise, this is done.

One note, I fixed what I believe to be a problem in Formation for this component. The padding on the hover state causes the sidenav item text to shift to the right. That's odd and unnecessary, possibly even distracting and unhelpful. I don't know if we have research to suggest otherwise but I'm inclined to not have the text shift. Where should I file an issue to address that CSS in VA.gov?

pdavies88:
@humancompanion-usds thanks for the updates. Let me take look into the Pattern issue and see what is going on there. In regards to the change to remove the text shift holistically I think the va.gov-team repo where we usually file tickets would be the best place.

pdavies88:
@humancompanion-usds I fixed the Patterns icons not appearing we were missing `status` on the index.md files for Forms and Templates. Please let me know if there are any other changes to be made or if this should be merged.

humancompanion-usds:
Okay, the fix to the maturity scale icons is working great. 

I did catch that somewhere along the way the Patterns > Forms > Service History page lost the dot between it's title and mime type and so was acting oddly. Fixed that (will push that up momentarily). 

I looked at it again and at the code and I'd like the behavior from VA.gov in regards to sub-pages. When the nav is quite long because you've expanded a large section and the page is long you can get in this state where it is hard to get to the bottom of the side nav to select the last item in the nav. It's also related to the fact that the nav is sticky. 

Thus I'd like to have us align to the production behavior of having an accordion that wraps the sub-pages, if possible and not too difficult:
Screen Shot 2022-05-13 at 4 07 39 PM

The parent page (e.g. components/form/) name is the accordion drawer title (e.g. "Form"). Clicking that just toggles the accordion to show or hide the sub-pages. It would not load a page itself. The parent page would get a hardcoded title of "Overview" (rather than repeating the drawer title) and be at the top of the list of sub-pages. That would get us closer to how production behaves. Let me know if that's reasonable.

pdavies88:
@humancompanion-usds I have adjusted to use the accordions for the side navigation. Please take a look and let me know if you would like any other adjustments.

humancompanion-usds:
@pdavies88 - We're almost there. 

* We've lost the maturity scale indicator now on the top-level item (whichever component is the Accordion drawer title). It could be on Overview instead if that's easier. If the indicator goes in the drawer face then the text (component name) should all line up vertically down the list of components (makes it easier to scan).
* I adjusted the padding as it was really thin on components and then extra wide for drawer faces. That makes sense in production because everything lives within a drawer. However, here it gives the drawers more prominence, which isn't what we want. I set the new values in a variable and then reused it in _sidenav.scss. (As an aside, I know 10px/1rem isn't in our Spacing Units; we should probably add it)
* The other remaining problem is that the "Overview" shows as the current whenever a sub-page is selected. It probably should not, unless it is in fact the current page.

Issue #776, opened 2022-04-28
Improve search results on design.va.gov


k80bowman:
## Description
Accessibility improvements to the design.va.gov:
- Search page should show descriptions (use intro-text from yml front-matter)
- Search results should use h2s for page titles

## Details
This comes from feedback we received during a user research session with Angela Fowler.
Check with Natalie and Matt about design of search results when updating page titles to h2s.

## Tasks
- [x] Add page descriptions to search page results
- [x] Update search page result page titles to h2s
- [x] Verify design of search results with Matt and/or Natalie

## Acceptance Criteria
- [x] Search page results include page descriptions
- [x] Page titles are now h2 elements.

PR #954, opened 2022-06-16
Add description to search results

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/776

### DIFF
--- a/src/assets/javascripts/vendor/search.js
+++ b/src/assets/javascripts/vendor/search.js
@@ -302,7 +302,7 @@ var _$src_8 = {};
     resultsContainer: null,
     json: [],
     success: Function.prototype,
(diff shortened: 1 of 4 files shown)


andresriveratoro:
@humancompanion-usds 

Here's how it looks using h2s for titles and some additional styling to match VA.gov.

Screen Shot 2022-06-16 at 13 02 01

humancompanion-usds:
@andresriveratoro - That looks great! Just a small change to make given that I [just wrote guidance around this](https://design.va.gov/components/card#when-to-use-caution), let's just make one large content area with a white background for the results to all sit in rather than each result in a card. I think the Card approach here creates visual noise we don't need. As long as we have the spacing right (12px margin on top, 32px margin on bottom) for each block of the results we should be fine. The typography will do the heavy lifting here. Thanks!!

andresriveratoro:
@humancompanion-usds How about this?

Screen Shot 2022-06-16 at 16 33 19

humancompanion-usds (approved):
Awesome! Thanks.

PR #998, opened 2022-06-24
Fixes Research link to include closed issues. Clarifies some Button g…

humancompanion-usds:
…uidance. Puts content label sections into accordions and proper HTML tables.


humancompanion-usds:
@k80bowman - I can walk you through all of these changes tomorrow. There is more in here than I originally intended but I can explain it all in person.

Issue #1011, opened 2022-06-30
Mobile-friendly, expandable alert - Documentation update


caw310:
## Configuring this issue
- [ ] Add issue to appropriate epic
- [ ] Add Design System component label (such as `va-alert`)
- [ ] Add `component-new` or `component-update` label, if applicable
- [ ] Complete sections below
- [ ] Delete this section once complete

## Description
Update documentation and guidance for Mobile-friendly, expandable alert on design.va.gov. When adding a new component, use the [component documentation template](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/master/src/_components/template.md).

## Details
[add links to any design documents]

[add link to Sketch component page]

Be sure to use the appropriate [maturity scale](https://design.va.gov/about/maturity-scale) option. If this is a new component that has not yet gone through Staging Review, use the "Use with Caution: Candidate" label.

## Tasks
- [x] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [x] Request documentation draft review
- [x] Once documentation is approved, publish to design.va.gov

## Acceptance Criteria
- [ ] Component documentation is published on design.va.gov

andresriveratoro:
The Storybook examples are in. Reassigning to @humancompanion-usds

humancompanion-usds:
Okay, this is one where development got out ahead of design and I wasn't paying careful enough attention to what was in here. Talked to Allison about this component. The use case the team tested for is for information we need to alert a user about when they reach a page or form flow. Thus unlike an Alert, it is not due to user action. Therefore we think the only valid variations are info and warning. Error and success should be handled by our standard Alert component. 

Thus we need to circle back around with the team that generated this component for us and ask them for valid use cases where they have used these other variations. If there are none then we need to remove them. If there are some then I need to review them and they may need to return to DSC.

Sorry, I should have caught this sooner. The documentation that was written for this for the "experiment" is largely copy and pasting from Alert which was inappropriate. Allison indicated this was one of the first experimental components. Thus it seems like the process was not yet very robust.

k80bowman:
This component is currently in use in two places:
- The [facility-locator application](https://github.com/department-of-veterans-affairs/vets-website/blob/d96726162e27fdfaa173102bd23b028c98c5915c/src/applications/facility-locator/components/search-results-items/common/LocationCovidStatus.jsx#L46-L58) uses it for an info alert
[...]

PR #1037, opened 2022-07-07
Update documentation for alert-expandable

andresriveratoro:
Part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1011

### DIFF
--- a/src/_components/alert-expandable.md
+++ b/src/_components/alert-expandable.md
@@ -13,13 +13,38 @@ anchors:
   - anchor: Research
 ---
 
-## Example
+## Examples
+### Informational alert
 
-![image](https://user-images.githubusercontent.com/4960080/127147819-12d07cf2-b357-4058-81d5-56f0caf764a2.png)
+Used to provide helpful information or something that warrants a user’s attention. Not used for negative consequences.
 
-![image](https://user-images.githubusercontent.com/4960080/127148001-517c85d7-e6d9-401d-81d7-1d3b8f1dbeb1.png)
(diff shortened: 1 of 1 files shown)


src/_components/alert-expandable.md
```diff
@@ -13,13 +13,38 @@ anchors:
...
+
+### Error alert
+
+Used when there is a problem or something destructive is about to occur.
+
+{% include storybook-preview.html story="components-va-alert-expandable--error" link_text="va-alert-expandable" %}
+
+### Continue alert
+
+{% include storybook-preview.html story="components-va-alert-expandable--continue" link_text="va-alert-expandable" %}
```
andresriveratoro, line 43: Do we know the use case for a continue alert?

humancompanion-usds: I think we may need to reach out to the team that created this component. I'm not sure what the use case would be for this component with the lock icon. It does seem specific enough that they probably had something in mind.

humancompanion-usds:
We need more guidance here. Once you add the Storybook examples you can send this my way. I'll just add it to the queue of other things we're doing around Alerts. There are some use cases of this in [Facilities](https://www.va.gov/north-texas-health-care/locations/plano-va-clinic/) that are useful which I'd like to highlight as examples in this documentation.

andresriveratoro:
> We need more guidance here. Once you add the Storybook examples you can send this my way. I'll just add it to the queue of other things we're doing around Alerts. There are some use cases of this in [Facilities](https://www.va.gov/north-texas-health-care/locations/plano-va-clinic/) that are useful which I'd like to highlight as examples in this documentation.

The Storybook examples are in. 👍

PR #1044, opened 2022-07-12
fix link to system error messages

k80bowman:
Fixes the link to the system error message page from the index error message page.

### DIFF
--- a/src/_content-style-guide/error-messages/index.md
+++ b/src/_content-style-guide/error-messages/index.md
@@ -19,7 +19,7 @@ Error messages mainly appear in interactive experiences such as tools, forms, an
 
 The messages dictionary holds the title, content, component, and state for each scenario of message on the site. The messages dictionary is broken up into 4 types of messages:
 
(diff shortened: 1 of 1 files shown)


bkjohnson (approved):

bkjohnson:
It may be good to look into [adding a broken link checker to CI](https://dsva.slack.com/archives/G01BJ3ESXL4/p1651163396247699) so we can prevent these kinds of things from happening.

humancompanion-usds:
@k80bowman - I fixed this in PR https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/1043 I made a note in the comment. 

@bkjohnson - This actually works locally which is why I didn't catch it. I'm guessing Jekyll's included server behaves differently than our deployed environment? I would love a broken link checker given the number of updates we've been making (and will continue to make).

bkjohnson:
I think you're right - I know s3 can be weird about URLs and pages.

PR #1061, opened 2022-07-19
trying to reorder the templates again

k80bowman:
Let's try numbers this time and see if that works better for re-ordering the templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE/_DST-component_documentation.md
+++ b/.github/ISSUE_TEMPLATE/_DST-component_documentation.md
@@ -1,5 +1,5 @@
 ---
-name: "_DST - Component documentation"
+name: "DST - Component documentation"
 about: INTERNAL DST USE ONLY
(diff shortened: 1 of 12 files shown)


.github/ISSUE_TEMPLATE/_DST-component_documentation.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: "_DST - Component documentation"
+name: "DST - Component documentation"
```
bkjohnson, line 2: Should the actual filename change to match this?

bkjohnson (approved):

.github/ISSUE_TEMPLATE/_DST-component_documentation.md
k80bowman: I'm not sure it matters much. I do like how having the underscore groups these templates in the file viewer and they're not in the middle of the external templates.

Issue #1051, opened 2022-07-13
Add guidance for the "one thing per page" design principle


humancompanion-usds:
# Feature Request

- [x] I’ve searched for any related issues and avoided creating a duplicate issue.

## This update is for:

- [ ] Content style guide
- [ ] Component
- [x] Pattern
- [ ] Utility
- [ ] Other

## What is the name?

One thing per page is a design principle. Check suitability or Check eligibility is a pattern. Essentially, we need to document a replacement for the Wizard pattern.

## What is the nature of this update?

- [ ] How to build this component/pattern
- [x] When to use this component/pattern 
- [x] When to use something else
- [ ] Usability guidance
- [x] Accessibility
- [x] Implementation
- [x] Research insights
- [ ] Package information

## What problem does this solve?

Progressive disclosure of form elements 

### Out in the world
* https://designnotes.blog.gov.uk/2015/07/03/one-thing-per-page/
* https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page
* https://beta.ssa.gov/prepare/check-eligibility-for-benefits
* https://home.treasury.gov/policy-issues/coronavirus/assistance-for-state-local-and-tribal-governments/emergency-rental-assistance-program/service-design/application-web-sites

### At VA
* https://dsva.slack.com/archives/C7NJWUA8K/p1648570003292579
* https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/appointments/va-online-scheduling/research/2021-request-clarification/Requests%20discovery%20-%20report.md
* https://va-scheduling.invisionapp.com/console/share/EW35S0RT8V/823651609
* https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/checkin/research/veteran-facing/mvp-usability/research-findings.md
* https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/checkin/research/veteran-facing/mvp-UAT/research-findings.md
* https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/checkin/research/veteran-facing/telephone/research-findings.md

humancompanion-usds:
## Steps to get to some usable patterns

* [ ] Clean up the current Patterns > Form > Overview page 
* [ ] Create new pattern: Patterns > Form > Sub-task
[...]

PR #1063, opened 2022-07-19
Start of work on creating a Wizard replacement. [Amends #1051]

humancompanion-usds:


humancompanion-usds:
@k80bowman - Hoping you can review this sometime this week and get the draft pattern out there. I realize it's not ideal to be talking about a pattern on other pages that is not yet published. Makes me wonder if we shouldn't make drafts visible (it's a function of Jekyll so we'd have to use some other mechanism to mark them).

@caw310 - Once it's out there I'd like feedback from the following which they can leave in issue #1051:

- Natalie
- Sean
- Allison
- Josh Kim
- Peter Russo
- Kristen McConnell

They all asked for this or contributed to it in some way. I'll do my best to incorporate their feedback when I get back and update the Wizard pattern to point folks to this pattern instead. It's a draft so you'll need the URL which should end up being: https://design.va.gov/patterns/forms/sub-task

Thank you both! Greatly appreciated. See you soon.

src/_patterns/forms/check-a-service-is-suitable.md
```diff
@@ -0,0 +1,68 @@
+---
+layout: pattern
+title: Check a service is suitable
```
k80bowman, line 3: Not a blocker, but could we re-title this "Check that a service is suitable"?

src/_patterns/forms/confirmation.md
```diff
@@ -26,6 +26,8 @@ The form confirmation page consists of:
 5. Action link  
 6. Helpful content  
 
+Examples of [confirmation pages in the wild can be seen here](https://dsva.slack.com/archives/C0NGDDXME/p1579042705077800)
```
k80bowman, line 29: Not a blocker, but we should probably not be linking to Slack on our public documentation site. Could we add links to examples here directly?

k80bowman (approved):

caw310:
I've reached out them to review and add comments to the original ticket. Thanks!

PR #1074, opened 2022-07-27
Update link component documentation

### LINKED ISSUE #976: Link - Documentation
## Description
Add or update documentation and guidance for Link on design.va.gov. When adding a new component, use the [component documentation template](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/master/src/_components/template.md).

## Details
[add links to any design documents]

[add link to Sketch component page]

Be sure to use the appropriate [maturity scale](https://design.va.gov/about/maturity-scale) option. If this is a new component that has not yet gone through Staging Review, use the "Use with Caution: Candidate" label.

## Tasks
- [ ] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [ ] Request documentation draft review
- [ ] Once documentation is approved, publish to design.va.gov

## Acceptance Criteria
- [ ] Component documentation is published on design.va.gov

k80bowman:
Closes #976 

Updates link component documentation to use examples from Storybook.

Also includes a small language change to the DST announcement issue template.


humancompanion-usds (changes requested):
For the component announcement a few things:

1. I'm not sure an issue template makes sense. Perhaps putting this at the top of "What's new" commented out? We have so many issue templates I'd rather we find a more appropriate location for this. 
2. The content is fine though we'd link to the component first and foremost.
3. Let's talk about a maturity scale adjustment to capture this state of creating a new web-component for an existing component that is in heavy use (like links and buttons). Perhaps "In QA" or "In Review" to reflect that it's going through the CollabCycle? It's going to continue to cause confusion if we make all new web-components Use with caution until the Cycle is done.

On the Link changes:
* I think the storybook example should come before the text. We don't seem to have a set pattern for this currently but I think that makes the most sense, to see the example and then read about it. Thoughts?
* For Download, Video, and Channel we can drop the lines that start "Structure". That should all be handled by the component now, yes?

k80bowman:
@humancompanion-usds I updated the link guidance, sorry I missed those "structure" items.

As for the announcement ticket, I agree that having an issue template seems like a bit much. It's there mainly so we can have some sample announcement text, but I wonder if maybe including this as part of the final review template would be better. I like the idea of having an announcement template commented out at the top of the What's New page, too. What do you think? I could remove this template and add "announce the component in Slack and on the What's New page" as a task in the final review ticket after the Staging review is complete.

For the maturity scale, yes, I would like to talk with you about this. While you were out, I discussed having a section in the Storybook sidebar for components "In development" with the engineering team and we all agreed this would work well. The idea was that we could use this section to house the web components we are building to replace React components during that period while they are being developed, before they go through Staging review. After Staging review, then, we can move them up to be the primary code example for that component and the React version can be deprecated. Could we use that same language in the maturity scale? So "Use with Caution: In development"? I would be OK with "In review" as well, if that works better for you. I would just like to coordinate the language we are using.

PR #1084, opened 2022-08-04
add small audit to design ticket

k80bowman:
Adds a small update to the component design ticket template for conducting a small audit when we are building a new version of a pre-existing component.


GnatalieH (commented):
Hi @k80bowman. Thanks for updating this template. I noticed a typo: alreday

I'm wondering if we should quantify the audit more specifically than "small". Sometimes these audits can be comprehensive, sometimes not, even if it's just updating a component. But for the sake of this template, how about something like:

"...conduct a small-scale audit (3-5 examples) to make sure there aren't design issues that need to be addressed."

k80bowman:
Thank you @GnatalieH! I've fixed the typo and used your language about the audit.

humancompanion-usds (approved):

PR #1091, opened 2022-08-09
Collects all of the banners together and updates their guidance. 

### LINKED ISSUE #816: Reconcile Full-width Alert vs. Banner
Full-width Alert: https://design.va.gov/components/alert#full-width
Banner: https://design.va.gov/components/banner

We're missing the Full-width Alert variation in our Sketch file and thus I was just about to add it. But I'm reading the guidance and I'm hard pressed to tell the difference between this Alert variation and the Banner. My understanding on the use cases for Banner from Dave Conlon was essentially what is under the Full-width Alert variation (i.e. a hurricane alert; government shutdown affecting VA services, etc.) They show up in different places but otherwise how does one choose between them?

Discussed with Allison who offered up this:

Per our guidance, full width...
* Full-width alerts are used only for emergency or urgent communications and should appear below the main navigation.
  * So would urgent be the blue info alert style? I assumed urgent would be the red or yellow. 

And our guidance for banner is...
* When there is very important information that should exist on the top of the webpage that should be visible to all users.
  * I assumed that would be blue info or possibly yellow warning. But not red
Emergency alert (full width) should appear below the main navigation and very important information (banner) at the top of the webpage.

It strikes me that there are two problems here:

1. The styles of alerts offered in both cases seem dubious.
2. Very important information vs. emergency information. The urgency of the comm is also relevant here though, again, we just decided to put the IE11 warning for those users in a Banner which will run for week or months and thus, in my mind, is not terribly urgent. Timely maybe.

In short, we need to do some investigation and make some decisions in order to provide clear guidance on when to choose. We may also decide to merge these into different variations of the same component.

humancompanion-usds:
Removes full-width alert. [Fixes #816] 

Updates what is new and adds a banner to it. Adds feedback link to the footer.


k80bowman (approved):
This looks good! I love the changes to What's New and the addition of the What's New announcement.

I have a couple of of questions, but they're not blocking.
1. Should we be using one of our components for the What's New announcement?
2. How do you feel about the `index` file for each sub-component in the sidebar automatically being labeled "Overview"? I'm not sure I care for that, for things like Button and Banner, we're not really giving an overview of banner components. It's documentation for the Banner component. I think it makes sense for Form, but not for the others.

humancompanion-usds:
@k80bowman - We don't have a component appropriate for the What's New slim banner. I don't want to give anyone the impression that we need yet another banner on the site. 

I'm not crazy about "Overview" but that was all Preston could sort out at the time. it should really just use the title of the page and we can customize the page titles (and keep the links the same by using the permalink param). Let me see if I can make that change quickly.

Issue #968, opened 2022-06-22
File Input web component - Documentation


caw310:
## Description
Add or update documentation and guidance for File Input on design.va.gov. When adding a new component, use the [component documentation template](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/master/src/_components/template.md).

## Details
Existing documentation on File Upload - https://design.va.gov/patterns/forms/file-upload
[add links to any design documents]
[add link to Sketch component page]

Be sure to use the appropriate [maturity scale](https://design.va.gov/about/maturity-scale) option. If this is a new component that has not yet gone through Staging Review, use the "Use with Caution: Candidate" label.

## Tasks
- [ ] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [ ] Request documentation draft review
- [ ] Once documentation is approved, publish to design.va.gov

## Acceptance Criteria
- [ ] Component documentation is published on design.va.gov

jamigibbs:
@humancompanion-usds @GnatalieH I've created a draft PR for the `va-file-input` component documentation here: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/1117

I've added the storybook examples, code usage, and a few other updates based on [this template](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/src/_components/template.md) but I think there might still be a few gaps related to design and UX feedback. I'm not sure if there is anything additional we want to add related to that.

PR #1117, opened 2022-08-18
File Input documentation updates for va-file-input web component

jamigibbs:
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/968 

Updates the documentation and guidance for File Input on design.va.gov for the `va-file-input` web component.

## Tasks
- [ ] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [ ] Request documentation draft review
- [ ] Once documentation is approved, publish to design.va.gov


humancompanion-usds:
I've contributed guidance updates but have one question for Beth, who is on vacation, around the error messaging for file input. We're using the [Alert error message](https://design.va.gov/content-style-guide/error-messages/feedback) in the inline error slot. I don't think we should be doing that. If that's the case, we may need to update Sketch and our Storybook example for error.

That said, I'd be okay with us pushing out what we have, as it's clearly better than what is there now, and then filing a new bug to figure out the error message. Alternatively, we can wait until Beth and Natalie are back next week.

humancompanion-usds (approved):

Issue #1129, opened 2022-08-24
Create shared search library


caw310:
## Description
There was a request for Experimental Design [Search Result Card] that was discussed in the 8/05/22 DSC meeting. 
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1085

Action items from that meeting were:
- Create template for Search shared library in Search 
- Set up time with Aslan to talk through how they create things in Sketch. 
- Document shared libraries 

## Tasks
- [x] Create template for Search shared library in Search
- [x] Set up time with Aslan to talk through how to create things in Sketch
- [x] Draft documentation on share libraries

## Acceptance Criteria
- [x] Shared search library created
- [x] Had meeting with Aslan to talk through how to create things in Sketch
- [x] Documentation written on shared libraries

GnatalieH:
@humancompanion-usds was the intention of this ticket to create a shared library with items that relate to search, only? I don't think I was in the meeting where this ticket was discussed. Because, there are a couple of potentially reusable patterns that stem from the [Experimental design request](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1085) created by @jcklpe, and one pattern is not directly related to search, rather the facility locations page.

**Example 1: facilities search results card**
![image](https://user-images.githubusercontent.com/88347129/188931596-36300a70-8e32-4d8a-98ef-93249d20e970.png)

I have created this asset in Sketch.

**Examples 2: facility locations page cards**
![image](https://user-images.githubusercontent.com/88347129/188931748-2553dc5a-f91e-40ee-be58-2c1e60e0cc96.png)

I have not yet created this asset because I wanted to make sure that we want to include it in the library, if the library is meant for search-specific items. This is more of a Facilities-specific pattern. It still could be useful for other Designers on the Facilities team.

humancompanion-usds:
@GnatalieH - Anything related to Search could be included, regardless of what kind of page it appears on. Facilities patterns are germane here as examples of search results on VA.gov. Aslan's request was the use case that generated this response so starting with those is a great first step.

GnatalieH:
Thanks @humancompanion-usds. I referenced the Experimental ticket, chatted with Aslan yesterday and got his clarification. I just wanted to make sure we were all on the same page. Thanks!

GnatalieH:
[...]

PR #1215, opened 2022-09-20
Fixes use of web components. [Amends #1129]

humancompanion-usds:


_config.yml
```diff
@@ -71,14 +71,13 @@ collections:
...
     permalink: /:collection/:title
 
-includes:
-  - /components/code
+include:
+  - vendor/javascripts/component-library/dist/esm/_commonjsHelpers-8b28c6fa.js
 
+# Exclusions are relative to the source directory (which is set to src)
 exclude:
-  - node_modules
```
bkjohnson, line 78: Why are we removing `node_modules` from being excluded?

humancompanion-usds: I added a comment above the exclude - it's relative to what we define as source. Thus it's only excluding things it finds in the src directory, which does not include node_modules. In short, that line was useless.

bkjohnson: Gotcha.

bkjohnson (approved):

Issue #1255, opened 2022-10-04
Fix the maturity scale indicators in the design.va.gov sidenav


k80bowman:
## Description
The way the sidenav on design.va.gov is built, the maturity scale indicators for component subpages represent the maturity scale for the parent page, rather than the child page. This means we end up in situations like the one pictured below where the sidenav indicator does not match the component maturity level.
![image](https://user-images.githubusercontent.com/12739849/193832604-8b8e5e25-d195-42cf-9aea-bed3dcffa3d5.png)

We also would like to, as much as possible, rely on a single source of truth for the maturity scale information. For our web components, this data is stored with the component in the `component-library` and is accessible through the `src/_data/component-docs.json` file, which fetches this information on site build. We need to, for our components that are web components, depend on this data for the maturity scale information, rather than maintaining it in the component yaml frontmatter.

## Tasks
- [ ] Fix the design.va.gov sidenav so that the component subpages have the correct maturity scale indicators
- [ ] Update the web component maturity scale indicators and tags to depend on the data in component-docs.json.

## Acceptance Criteria
- [ ] The design.va.gov sidenav has the correct maturity scale indicators for component subpages
- [ ] There is a single source of truth for the web component maturity scale statuses

humancompanion-usds:
I accidentally removed status indicators for patterns.

PR #1306, opened 2022-10-26
Fixes maturity level indicator in sidenav for sub-pages. [Amends #1255]

humancompanion-usds:
I wanted to share this to show my work and to demonstrate how using the component-docs.json file is just not going to work for a number of reasons. Thus I'm going to have to refactor this to use the front matter in Jekyll instead. Here are the problems:

1. Maturity naming. We took some liberties in the naming of the maturityCateogry and maturityLevel. For example, "caution" instead of "use-with-caution" and "best_practice" instead of "best-practice". This isn't so hard to overcome and I did so in this code but, it's annoying.
2. Component naming. We also are inconsistent here and thus trying to take the title of the page and using slugify and append to come up with the component name is not working for some components. 
  * Sometimes it is just an abbreviation of the full title: "Radio button" is just "va-radio", "Date input" is "va-date", etc. 
  * Other times it is because of the way I re-wrote the titles: "Banner - Promo" is "var-promo-banner", "Progress bar - Segmented" is "va-segmented-progress-bar".
4. Components that are not web-components. We actually have a fair number of items that are not web-components yet:
  * Address block
  * MaintenanceBanner
  * Card
  * Divider
  * Form > Input message
  * Form > Label
  * Link > Action link
  * Link > Collection
  * Link > Deep Content
  * Sidenav
  * Tabs
  * Tag
  
Thus for now Jekyll front-matter is going to be the source of truth for maturity status until we can resolve this issues.

### DIFF
--- a/src/_components/form/date-input.md
+++ b/src/_components/form/date-input.md
@@ -2,7 +2,7 @@
 layout: component
 permalink: /components/form/date-input
 has-parent: /components/form/
-title: Date Input
+title: Date input
 research-title: Form controls
 intro-text: "Use the date input component to help users enter a date they would know or a date they can approximate."
 status: use-deployed
--- a/src/_components/form/memorable-date.md
+++ b/src/_components/form/memorable-date.md
@@ -2,7 +2,7 @@
 layout: component
 permalink: /components/form/memorable-date
 has-parent: /components/form/
-title: Memorable Date
+title: Memorable date
 research-title: Memorable Date
 intro-text: "Three text fields are the easiest way for users to enter most dates."
 status: use-with-caution-available
--- a/src/_components/form/radio-button.md
+++ b/src/_components/form/radio-button.md
@@ -2,7 +2,7 @@
 layout: component
 permalink: /components/form/radio-button
 has-parent: /components/form/
-title: Radio Button
+title: Radio button
 research-title: Form controls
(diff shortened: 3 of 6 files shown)


jamigibbs:
@humancompanion-usds Should a follow-up ticket be created for this to figure out a way to put the source as the component-library (component-docs.json or whatever is needed to centralize it there)?

jamigibbs (commented):
Should button and link be green?

Screen Shot 2022-10-27 at 12 40 27 PM

Screen Shot 2022-10-27 at 12 42 14 PM

humancompanion-usds:
I'm actually fine with using the Jekyll front matter because the more I think about this the more I'm inclined to believe that we'll always have things that are not yet web-components where we'll want to show their maturity level. We will likely also continue to have things like Address block that we won't bother to convert to a web-component. I suppose we could try to find a hybrid way to have some state come from component-library and the rest come from YAML? I just don't know that it is worth the complexity and it certainly isn't a priority.

The maturity level indicator is more about the how long the component has been existed in the design and implementation of VA.gov rather than how long it's been a web-component. Katy and I never really synced on that understanding, but that's the intent. So at this point Button and Link should be green in that we've not altered how those are designed and implemented even though the code implementation is different in that they were achieved via HTML and CSS and now they are achieved via a web-component (which is largely just "codifying" that same HTML and CSS).

jamigibbs (approved):

Issue #1298, opened 2022-10-20
Update Telephone to include SMS - Documentation


jamigibbs:
## Description
Once [the SMS option has been added](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1246) to the va-telephone component, we should update the guidance for Telephone on design.va.gov to include details for SMS.

## Details
- https://design.va.gov/components/telephone
- https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/a/WKjQYn0

## Tasks
- [ ] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [ ] Request documentation draft review
- [ ] Once documentation is approved, publish to design.va.gov

## Acceptance Criteria
- [ ] Component documentation is published on design.va.gov

humancompanion-usds:
Confirmed that all docs are now up-to-date with the SMS option.

PR #1321, opened 2022-11-03
Updates guidance for Telephone and shared content guidance between th…

humancompanion-usds:
…is page and the Dates, times, phone numbers, and addresses page. [Amends #1298]

### DIFF
--- a/src/_components/telephone.md
+++ b/src/_components/telephone.md
@@ -8,6 +8,7 @@ anchors:
   - anchor: Examples
   - anchor: Usage
   - anchor: Code usage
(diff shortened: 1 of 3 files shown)


humancompanion-usds:
I'm waiting on a review from Beth and/or Danielle on some content consideration changes I made but this is otherwise done.

jamigibbs (approved):
This looks good to me but I did notice that locally the sms prop was not getting pulled through here. This might just be something going with my local setup though.

Screen Shot 2022-11-04 at 11 20 05 AM

PR #1385, opened 2022-12-06
New pattern 1109

humancompanion-usds:


GnatalieH (approved):
This looks good to me. The only thing I can think of is adding a "When not to use this pattern" section in the Addresses pattern and mentioning the Address block component. Address block should only be used for read-only addresses, locations Veterans can visit or send correspondence to.

PR #1405, opened 2022-12-16
Update multiple responses pattern 1147

humancompanion-usds:
Updated list and loop to discuss all 4 options. More to come on this one.


GnatalieH (commented):
@humancompanion-usds here are my thoughts:

- The "What to watch out for" section under Single page is an H4. Is that on purpose? I also am not sure what a Designer or Developer, for example, would do with this information. There isn't guidance other than to watch out for the error scenario. How would someone mitigate this error scenario if they wanted to use this pattern?
- Under Add item the second bullet point should say "displayed" rather than The "Add a new [item]" is display..."

Otherwise looks good! Thanks.

GnatalieH (approved):
Approved, thanks!

PR #1413, opened 2022-12-27
Update email-address.md

taylojill:
Hello @GnatalieH and @humancompanion-usds. I made the following updates to the documentation. Please review and provide feedback. Thanks!
1. Added 2 screen shots (prefilled email address and prefill information)
2. Under How this pattern works, added notes to: validate email addresses, inform users about prefilled information, and that changes to the form will not update their VA.gov profile.
3. I assumed that the prefilled email address field could be overwritten by the user, so I updated the error message templates.


GnatalieH (approved):
Hi @taylojill. This looks good to me. Thanks.

GnatalieH:
@taylojill I want to make sure about that the prefilled email address can be overwritten before I merge this PR. I would assume that is the case, too.

PR #1424, opened 2023-01-10
Update confirmation.md

taylojill:
Hello @GnatalieH and @humancompanion-usds I updated the "Structure" section based on content from the Confirmation guidance page in GitHub. Please review and provide feedback. Thanks!
1. Added a sentence to the H1 section.
2. Added an example to "Success alert" section.
3. Added an example to "Print this page pattern" section. I also updated the blue box section.
4. Added an example to the "Other content elements" section.

### DIFF
--- a/src/_templates/forms/confirmation.md
+++ b/src/_templates/forms/confirmation.md
@@ -35,7 +35,7 @@ The form confirmation page consists of:
 
 ### H1
 
-File for [benefit] or Apply for [benefit] and the VA paper form equivalency appears next to or below the H1.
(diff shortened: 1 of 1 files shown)


GnatalieH:
@taylojill I think these are great additions! I think we can remove the Related section and GitHub link when these recent updates are approved, as the content should totally overlap. CC @humancompanion-usds

humancompanion-usds (approved):
Agreed with Natalie. Please remove the related section and link to github and then approved! Thanks.

PR #1425, opened 2023-01-11
Create header-simplified

kristenmcconnell:
For review @humancompanion-usds.


Andrew565:
@humancompanion-usds Do we still need/want to merge this PR in?

humancompanion-usds:
Yes, I need to merge it but I also need to add to it and make some related changes. I also need to check back with the team on the progress.

@kristenmcconnell - Did this more minimal header get used and or tested yet?

kristenmcconnell:
@benbrasso-agile6 Can you add more detail about the simplified header for @humancompanion-usds?

benbrasso-agile6:
Check-in incrementally tested the simplified header and footer from 2/8 - 2/20. We looked at page drop offs and clicks on the VA logo (indicator of wayfinding somewhere else) to measure performance. On 2/21, we increased the rollout to 100% of Veterans and have been using it ever since. 

The form title could not be added into the header because that code is owned by sitewide public websites. If I recall correctly, we weren't able to change the VA header component itself (because it doesn't exist). Rather, we edited what we could based on what we had access to.

benbrasso-agile6:
@humancompanion-usds, I was trying to track down where we were with the simplified header component. Was there a benefits team working on it last? I know the check in team was looking to re-hide the "Sign in" link because we don't know how it got re-added. (Screenshot below.) And, I believe a benefits team was looking to see how what's in place could be altered to include a form title in the header. Although, I believe the sitewide team owns that code. Anyways, was trying to remember where this work was left off. Thanks! cc @cgednalske 
 
screencapture-staging-va-gov-health-care-appointment-pre-check-in-error-2023-10-02-11_42_46

humancompanion-usds:
The Veteran-facing Forms team is testing a minimal header in a form flow that has a number of other changes as well. They are half way through usability testing.

benbrasso-agile6:
TY

humancompanion-usds (approved):

PR #1429, opened 2023-01-13
Enabling multi-environment deploys

Andrew565:
This PR modifies the "Deploy site" workflow in GitHub Actions. 

Previously, whenever there was a **push** to the `main` branch, this workflow would build a deploy a copy of the updated code to all three "environments" - dev, staging, and prod. Every one, every time.

This PR changes the workflow to instead only push to one of the environments, depending on which branch was pushed to. Respectively, pushing to the `dev`, `staging`, or `main` branches will cause this workflow to build and deploy what was pushed to `dev-design.va.gov`, `staging-design.va.gov`, and `design.va.gov`. It accomplishes this by utilizing new "repository variables" that were created using the interface at Settings -> Security -> Secrets and Variables -> Actions -> Variables (screenshot of said variables below).

Screenshot 2023-01-13 at 8 32 10 AM

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,62 +3,62 @@ on:
   push:
     branches:
       - main
+      - staging
+      - dev
 
 jobs:
   deploy:
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


humancompanion-usds (approved):
I believe we're find to proceed with this. I would like to understand the "long way" though so that I can avoid situations like this PR where my changes ended up in this PR.

PR #1430, opened 2023-01-13
Update confirmation.md

GnatalieH:
I replaced the png images of example messages with block quotes. The pngs looked very small on the published page. @taylojill is OK with me making these changes. CC @humancompanion-usds


taylojill (approved):
Thanks @GnatalieH

PR #1451, opened 2023-01-20
Example update to homepage

Andrew565:
Just a quick example of updating the homepage.


Andrew565:
Ignore this, for docs purposes.

PR #1464, opened 2023-01-23
Create keep-a-record-of-submitted-information

taylojill:
Adding a draft for new Keep a record pattern.


GnatalieH (commented):
@humancompanion-usds tagging you for review.

humancompanion-usds (approved):
@taylojill - I fixed up a few things that we can go over. First, the file has to end in .md so that Jekyll knows that it is a Markdown file otherwise it won't compile. Second, in the status line in the YAML frontmatter you can't use a colon. Thus the statuses use dashes (so "use-deployed" instead of "Use: Deployed"). This too can keep the page from being displayed because Jekyll can't parse and compile it. 

The other changes were related to using more of the Pattern template to list out the components used in the pattern and connecting up the Sketch templates. We can discuss all of this so you know where to find this stuff. Thanks very much for putting this together. I'm happy to get this pattern out the door!

taylojill:
@humancompanion-usds - Thank you!

PR #1482, opened 2023-01-27
Adds Section links to Link - Collection. [Fixes #943] Cleans up some …

### LINKED ISSUE #943: Update/addition to link collection pattern
## What
In My VA, we use a collection of links and icons to highlight links that can take users to content that is related to an overall topic. The links themselves are not necessarily related to each other.  For example, under health care, a user has a health-care related set of links to this content:
- secure messaging
- manage Rx
- view test results
- etc

## Why
The My VA collection of links provides a way to have a collection without a heading, container, and with more lightweight borders than some of the examples on [the collection page](https://design.va.gov/components/link/collection). This pattern is useful for link-heavy pages such as My VA, that also have many headings to delineate sections. The icons and borders help differentiate the links in the collection from the links in the cards and alerts on the page.

This also offers different placement opportunities than the current guidance that suggestions the link collection be placed at the bottom of the page. Our link collection is used to provide the links in the context of a section of the page without distracting from the main task or call to action.

**What evidence do you have that it's needed by multiple services across VA?**
I don't have any strong evidence that it's NEEDED, but I know there are many use cases on content pages where we need to surface a variety of links to Veterans in a way that is meaningful, and easy to find, but not distracting from the main body.

**What evidence do you have that it meets the needs of the users of those services?**
I've heard many Veterans complain about a wall of links and struggling to parse through content heavy pages on VA.gov.

**Have you checked that it doesn't already exist in the VA.gov Design System?** 
This particular pattern doesn't exist but there are variations on [the collection page](https://design.va.gov/components/link/collection)

## Anything else
**Mock-ups**
[My VA Desktop](https://www.sketch.com/s/9b0e6efc-423a-4354-9db3-ab2083d566c9/a/AxrLJlM)
[My VA Mobile](https://www.sketch.com/s/9b0e6efc-423a-4354-9db3-ab2083d566c9/a/agw4D9Z)

humancompanion-usds:
…link page.


GnatalieH (approved):
Looks good, the only thing I'd consider is changing this text under the Choosing between variations header:

Review the when to use a collection section for guidance.

Change to:

Review When to use a Collection for guidance. 

The mention of "section" in the current text is confusing because we have a variant called "Section links" now.

PR #1524, opened 2023-02-08
Update component-library to version 13.10.1

jamigibbs:
This update was confirmed running locally using Node v18.14.0 and NPM v6.14.8:

Screenshot 2023-02-08 at 4 19 58 PM

Screenshot 2023-02-08 at 4 20 45 PM


Andrew565 (approved):
Verified this as working, as well.

PR #1525, opened 2023-02-08
Update readme with latest node version tested and add npm to package.json engine. 

jamigibbs:
We ran into some errors when trying to run this repo locally that suggested there was a problem with a React dependency. But I believe the problem was that the NPM version installed locally was not compatible with this repo. 

This PR adds NPM 6.14.8 to the [package.json engine](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#engines) so a warning should display if someone is trying to run this with an incompatible npm version. And it also updates the readme to show that it has been tested also in the latest version of node (18.14.0).

If someone needs to run this repo locally and they are on an npm version that is not compatible, they will need to run: 
`npm install -g npm@6.14.8`.


Andrew565 (approved):
LGTM

Issue #1434, opened 2023-01-18
Update Radio Button guidance with descriptive text and tile variations


caw310:
## Description
There are teams interested in using the Tile version of Radio button. It's only in Storybook right now so we need to add guidance to design.va.gov on use of the tile and description text for Radio button. 

## Details
Storybook links:
* https://design.va.gov/storybook/?path=/docs/components-va-radio--default#with-description-text 
* https://design.va.gov/storybook/?path=/docs/components-va-radio--default#tiled

## Tasks
- [ ] Write guidance for Radio button with tile and description text 
- [ ] Post on staging-va.gov for review and sign off

## Acceptance Criteria
- [ ] Radio button on design.va.gov has guidance for variation with tile and description text

PR #1552, opened 2023-02-14
Adds variations of Radio button and syncs guidance with USWDS. [Fixes…

humancompanion-usds:
… #1434]

### DIFF
--- a/src/_components/form/radio-button.md
+++ b/src/_components/form/radio-button.md
@@ -7,6 +7,7 @@ research-title: Form controls
 intro-text: "Radio buttons allow users to see all available choices at once and select exactly one option."
 status: use-deployed
 anchors:
+  - anchor: Examples
   - anchor: Usage
   - anchor: How to use
   - anchor: Code usage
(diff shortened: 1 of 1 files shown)


src/_components/form/radio-button.md
```diff
@@ -7,47 +7,80 @@ research-title: Form controls
...
-{% include storybook-preview.html height="200px" story="components-va-radio--default" link_text="va-radio" %}
+## Examples
+
+### Default
+
+{% include storybook-preview.html story="components-va-radio--default" link_text="va-radio" %}
+
+### Hint text
+
+{% include storybook-preview.html story="components-va-radio--with-hint-text" link_text="va-radio" %}
```
jamigibbs, line 26: I don't think that all of the component examples are doing this but I just wanted to make note that `link_text` should probably be more descriptive so that it's specific to the example it's linking to:

"va-radio with hint text"
"va-radio label header"
"va-radio with description text" 
etc....

Otherwise LGTM!

jamigibbs (approved):

PR #1557, opened 2023-02-15
Update component library to v13.10.3

jamigibbs:
Verified compiles locally:

Screenshot 2023-02-15 at 12 30 22 PM

And also displaying updated prop name from v13.10.3:

Screenshot 2023-02-15 at 12 30 32 PM


Andrew565 (approved):
LGTM

Issue #1658, opened 2023-03-31
Deprecation Phase 3 - Table React Component


caw310:
## Description
We will deprecate Table React component.

With the completion of the [Phase 2 ticket]([Deprecation Phase 2 of React components #1637](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1637)), we will need to move the Table React component into [Phase 3](https://vfs.atlassian.net/wiki/spaces/DST/pages/2243854357/Plan+-+Deprecation+plan+for+React+components#General-deprecation-strategy). This means that the following steps will need to happen:

- Begin migrating any remaining instances of the React component ourselves. 
- The pace of this work will depend on the team’s availability and the priority of this work over other planned work and obligations.
- Once there are no more uses, remove the React component from the component-library repository

## Tasks
- [x] Check if the component is still being used in vets-website
- [x] If it is...
    - [x] Reach out to teams to check if they have an update pending.
    - [x] Migrate any remaining instances of the React component ourselves. 
    - [x] Once there are no more uses, remove the React component from the component-library repository
    - [x] Remove the ESLint rule
    - [x] Remove it from the [vets-website GA script](https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/platform/site-wide/component-library-analytics-setup.js#L67-L73).
- [ ] If it is not...
    - [ ] Remove the React component from the component-library repository
    - [ ] Remove the ESLint rule

## Acceptance Criteria
- [ ] Deprecated component has been removed entirely.

jamigibbs:
The `Table` React component has a migration script available. When there are no more instances of the React component, also complete the following cleanup steps:

- [ ] Remove the `table-migration` from the web-components package: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/index.ts#L5-L6
- [ ] Also remove `va-table` from the migration list [on design.va.gov](https://design.va.gov/about/developers/using-web-components#how-to-migrate-to-web-components).

caw310:
This will carry over into the next sprint as there are more instances of the REACT table component than we thought.

caw310:
This will be put into our backlog for an upcoming sprint as there is an issue that needs to be fixed with va-table before the REACT component can be deprecated. 
Need this ticket done first. https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1821

PR #1868, opened 2023-06-22
Remove table from web component migration list

jamigibbs:
The React Table component has been deprecated and removed so I am removing it from the [migration list](https://design.va.gov/about/developers/using-web-components#web-component-migrations).

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1658


Andrew565 (approved):
LGTM

PR #1953, opened 2023-07-31
Adds guidance on linking to external sites. [Fixes #1680] [Fixes #1600]

### LINKED ISSUE #1600: Request update to documentation for Link component
### Duplicate check

### This update is for:

Component

### What is the name?

Link

- [ ] When to use this component/pattern
- [X] Usability guidance
- [X] Accessibility
- [ ] Implementation
- [ ] Research insights
- [ ] Package information

Would like to expand documentation (in behavior section) for this component in two ways: 1.) address expected behaviors for external links (both federal site and non-federal sites) with accessibility notes and 2.) more explicitly state that links to any pages within Va.gov should open within the same window (with exception that is listed in the DS now).

This issue came up recently when governance team posted an issue regarding this component at staging review. In that scenario, a team's pages were linking out to other VA pages - in new tabs. While that team may have just missed these ones, we realized that design system guidance is not explicit that VA.gov pages should always open with the same window. Further, additional links in the product that **do** link externally to Myhealthevet were missing any accessibility notations that would announce this to a screen reader. There is no guidance about this (or non-federal external links) in the DS.

### Additional Context

It looks like there is actually an OMB policy around this - see link from [USWDS](https://designsystem.digital.gov/components/link/) that addresses federal & non-federal external links.

### LINKED ISSUE #1680: Add section on linking to external sites on Component - Link and Content - Links
## Description
We need a sub-section on both [Component - Link](https://design.va.gov/components/link/) and [Content - LInks](https://design.va.gov/content-style-guide/links) (they already share content through a partial) for "Linking to external sites" that provides clear guidance.

## Details
Josh provided the following feedback to a recent request on this matter:

> There’s a couple ways of going about doing this, which may depend on y’alls context:
> * Including text within the link that indicates where they will go (e.g. Chronic B-cell leukemia on VA's public health site) But this might be a bit verbose for the context you’re handling if it’s repeated for every single link in the list
> * [USWDS uses an external link icon](https://designsystem.digital.gov/components/link/) (but gov uk debates its use here and removed it from their system due to the icon possibly causing confusion for sighted users)
> * Including text around those links that give the user a heads up that they will navigate to other websites, and consider using `aria-describedby` to connect them together
> * Use `aria-label` to create a screen reader only label that appends where they link out to
> * Provide a list of items that aren’t linked up, with one external link. For example:
>  * List of Cancers
> * `Learn more about these cancers on the VA public health website`

I concur that the external link icon is more trouble than it is worth. The USWDS just says to use it consistently. It does not say that it has to be used. We don't use it in the footer today and I'm not inclined to start. I agree with the UK.gov folks that it adds noise and is a solution in search of a problem. That said, our position from the perspective of the Design System is unclear and we should provide guidance on this topic.

## Tasks
- [ ] Write guidance
- [ ] Publish
- [ ] Get a review from CAIA 

## Acceptance Criteria
- [ ] Guidance on linking to external sites is provided on design.va.gov on the Component - Link and Content - Links pages.

humancompanion-usds:


jamigibbs (approved):

src/_components/link/index.md
```diff
@@ -106,9 +106,10 @@ Review ["Usage"](#usage) for guidance.
 
 ## Accessibility considerations
 
-- **Material honesty**: Do not style a link to look or behave like a button ([material honesty](https://alistapart.com/article/material-honesty-on-the-web/)).
-- **Keyboard navigation**: The user must be able to navigate to links using the Tab key and activate links using the Enter key.
-- **Purpose and target**: Without a purpose and target, everyone-- but especially screen reader users-- will struggle to understand where they may be routed off to.
```
megzehn, line 111: Could we clarify this language?

Maybe: **Purpose and target**: Link text that doesn't indicate a clear purpose or destination makes it harder for everyone--especially screen reader users--to understand where they're getting routed off to.

PR #2158, opened 2023-10-04
Adds a page on creating components in Figma. [Fixes 2119]

humancompanion-usds:


src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+* Properties should be listed in an order that mimics the component’s visual layout from top to bottom and left to right.
+* For main components, show nested properties as appropriate.
+* Properties should be named and set up consistently:
+  * Mode: Always appears first. Some options include Light and Dark or languages.
+  * Type: Appears below Mode. Used for things like Primary / Secondary or H1 / H2.
+  * Layout: Appears below Type. Used for things like Horizontal / Vertical or varying quantities.
+  * State: Appears below Layout. Used for things like Focus / Active or Open / Closed or Active / Inactive.
+* Booleans should be named and set up consistently.
+  * Used to show/hide icons, buttons, content, etc.
+  * Always name them descriptively – i.e. icon (left) / icon (right), button (secondary), header (show/hide).
```
danbrady, line 51: Some of these examples might cause confusion. Perhaps the boolean property naming should follow an "[Action] [Noun]" format. Also, to be consistent, perhaps the Action should be consistent to follow the "true" value. For example, a label should be "Show Icon", not "Hide Icon" or "Icon (Show/Hide)".

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+* Instance swaps should be named and set up consistently.
+  * Preferred values should be set as appropriate.
+  * Always name them descriptively – i.e. icon (left) / icon (right).
+* Text properties should be named and set up consistently.
+  * Add content fields for every editable content area.
+  * Header and title fields should be named descriptively – i.e. header / title or header 1, header 2, etc.
+  * Body copy fields should always be named “content”.
```
danbrady, line 58: Can we agree to use Title Casing in the naming of everything? I.e. "Content", "Title", "Show Icon".

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+
+### States
+
+The following states should be depicted on all components:
+
+* Default
+* Hover
+* Active
+* Focus
+
```
danbrady, line 70: I've noticed some Sketch elements do not have some of these states. For example, the checkbox doesn't have a hover. Initially, should we default to what Sketch has for the first pass and circle back to fill in after all the components are in Figma?

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+### Sub-components
+
+For any sub-components that are used to build another component that should NOT be available to a designer using the library, prefix the component with an underscore or period (i.e “_” or “.”). This will make those components “private” to the current Figma file. It will not be exposed for a designer to use on it’s own with connected to the library.
```
danbrady, line 73: "...*when* connected to the library", not "with".

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+## Design tokens
+
+All components should be connected to the VADS Component Library styles & variables.
+
+* All components should be connected to the proper style or variable/token 
+* **Colors:** Connected to the correct color token (not primitive!) found in the local variables library.
+* **Typography:** Connected to the correct text style found under local styles.
+* **Spacing:** Connected to the correct spacing token or primitive. 
+* **Icons:** Connected to correct icon found on foundation page.
+
```
danbrady, line 90: When a token is applied to a value, the value gets a colored background to differentiate it:
 
Screenshot 2023-10-05 at 1 35 17 PM
Screenshot 2023-10-05 at 1 35 29 PM

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+### Spacing
+
+* **Apply spacing variables whenever possible.** Look for the logo below to access spacing variables. 
+
+{% include component-example.html alt="Figma spacing logo." file="/images/about/designers/figma/spacing-logo.png" caption="Spacing variables can be found with auto layout, radius, and width and height." %}
+
+### Iconography
+
+* **Only use icons from the VA Design System.** Currently, these are [Font Awesome version 5.15.4 icons]({{ site.baseurl }}/foundation/icons) however, we are moving to the [USWDS icon set](https://designsystem.digital.gov/components/icon/). Some USWDS v3-based components are already using USWDS icons.
```
danbrady, line 122: I've just learned there are Font Awesome, USWDS, and Material icons all in use. I'm guessing we'll need to bring them all in, and eventually deprecate the Font Awesome ones? Also, how do we manage icons when USWDS updates their set?

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Quality assurance
+
+There are multiple levels of quality assurance performed on components:
+
+1. **Test of the design of the component in a design tool.** The Design System Team tests components built in Figma (and Sketch, UXPin, et. al)
+2. **Test of the built web component.** The Design System Team then tests the component again once it is built as a web-component.
+3. **Governance review of the component.** Finally, the Governance team tests and reviews the component for design, content, information architecture, and accessibility. They also review the design assets and guidance on design.va.gov. 
+
+Some of the tests, such as color contrast checks, that the Design System Team performs will be re-tested by the Governance team in the built component. Thus ideally designers will perform these tests themselves when creating a component.
```
danbrady, line 132: Perhaps we should add a line to the Colors section about testing for color contrast issues. (The "Contrast" plugin is helpful for an initial pass.)

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+
+- [ ] Does the component naming make sense and follow Storybook (if the component has been built)?
+- [ ] Does the component link to guidance in design.va.gov?
+
+#### Auto layout
+
+* Components should be as reactive as possible. Most, if not all, components should be able to be resize for either mobile, tablet, or desktop viewport widths.
+* The correct spacing should be applied to auto layout.
+* Try to break the component! 
+  * Test on asset NOT parent component
```
danbrady, line 153: I think we're saying "instance" instead of "asset" and "main component" instead of "parent component" to use Figma language. However, a designer cannot edit the main component within a design file that subscribes to the DS library. So this line might not be needed.

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+
+* Ensure any [sub-components](#sub-components) are named correctly.
+
+#### Variations
+
+* If the component already exists in the system, ensure all variations match what is built in [Storybook](https://design.va.gov/storybook/).
+
+#### Tokens and variables
+
+* All components should be connected to VADS Component Library styles & variables .
```
danbrady, line 183: Remove extra space before period. All our tokens have a "vads-" prefix, right? We should mention that: "Ensure all variables used have the 'vads-' prefix.

src/_about/designers/creating-components.md
```diff
@@ -0,0 +1,188 @@
...
+* All components should be connected to the proper style or variable/token:
+  * **Colors:** Connected to the correct color token (not primitive!) - can be found in local variables library.
+  * **Typography:** Connected to the correct text style - under local styles.
+  * **Spacing:** Connected to the correct spacing token or primitive.
```
danbrady, line 187: I don't think we should be connecting to a primitive here.

danbrady (changes requested):
I left a few inline comments. Happy to discuss...

src/_about/designers/creating-components.md
humancompanion-usds: No, I don't want to propagate bugs from Sketch to Figma. I also don't want to be iterating on the Figma library forever. Part of the reason we are documenting this stuff is to ensure that we actually do it. If there are legit exceptions that is fine.

src/_about/designers/creating-components.md
```diff
@@ -95,6 +100,10 @@ All components should be connected to the VADS Component Library styles & variab
 
 All colors should be pulled from the variables in the VADS Figma file.
 
+* **All color combinations much pass color contrast at the following sizes and ratios.**
```
danbrady, line 103: "must" instead of "much"

danbrady (changes requested):
Looks good. Just a typo and we're GTG.

src/_about/designers/creating-components.md
humancompanion-usds: We talked about this issue this afternoon. Carol recorded the meeting for you. The short story is that the v3 versions of the MVP components do, mostly, use the USWDS icons. There are a few bugs there (notably Accordion) where we didn't switch to the USWDS icon. Material icons were only brought in where there was no equivalent to a USWDS icon. USWDS icons are actually Material underneath so they advised us to do that.

humancompanion-usds: When USWDS updates their set we'll pick up those updates. They don't maintain a Figma library but, if ours is good, we could gift it to them.

danbrady (approved):

PR #2224, opened 2023-10-23
Init export of Figma variables [Closes #2155]

### LINKED ISSUE #2155: Export Figma Variables to JSON
## Description
In order to have a single source of truth for tokens, the team decided that Figma should be that place. Currently, we do not have a way of exporting the native variables set up in Figma to a format that could be used or referenced by multiple teams or platforms. 

Figma does not have a native way to export variables (but I'd bet it's likely to happen). So we'll have to rely on a plugin to do the export. Two that we found so far to try are:

- [Design Tokens](https://www.figma.com/community/plugin/888356646278934516/design-tokens) - To be explored first
- [Export/Import Tokens](https://www.figma.com/community/plugin/1256972111705530093/export-import-variables)
- [Sheets to variables](https://www.figma.com/community/plugin/1255679829738098695/sheets-to-variables)

## Tasks
- [x] Explore plugins and validate successful export of JSON

## Acceptance Criteria
- [x] Validate variable name, value, description, and any other information is successfully exported

cc @humancompanion-usds @micahchiang @caw310

danbrady:
# Figma export of tokens to JSON #2155

- Uses [Design Tokens](https://www.figma.com/community/plugin/888356646278934516/design-tokens) plugin
- Follows the current W3C draft proposal for JSON token format
- Exports all variable/token types from all collections (not hidden by a "_" or "." prefix)
- Exports description
- Exports features that can not currently use Figma variables (typography, grids, etc)
- Supports alias values to other tokens
- Supports additional token information (how tokens are scoped)

How to run the plugin:
  1. While in the component library Figma file, find the Design Tokens plugin:
  Design Tokens plugin

  2. Upon clicking "Run", a context menu will popup. Select "Settings".

  3. The following settings were used for this initial export:
   Plugin Settings window

  4. Select "Save Changes"

  5. Run the plugin again choosing "Export Design Token File" with the following settings:
    Export settings
    
  7. PR the exported JSON to `src/_data/tokens/vads.tokens.json`


Andrew565:
Looking at this output, I'm concerned that we will need to create/employ a transformer script of some kind in order to get this data into a format that would be usable by style-dictionary. As an example:

```
"vads primitives": {
    "color": {
      "base": {
        "vads-color-white": {
          "type": "color",
          "value": "#ffffffff",
          "blendMode": "normal",
          ...
          }
        },
```
If this were put into style-dictionary directly, the output variable would be `vads primitives-color-base-vads-color-white`, if it didn't just break at the space in the name. So we would need to transform this into something more like:

```
"color": {
  "base": {
    "white": {
      "value": "#ffffffff",
    }
  }
}
```
This is, of course, doable. Just making a note that extra steps will be required to take this output and make it usable by style-dictionary.

danbrady:
> If this were put into style-dictionary directly, the output variable would be `vads primitives-color-base-vads-color-white`, if it didn't just break at the space in the name.

I see what you mean, @Andrew565. I'm surprised that the plugin doesn't automatically escape/convert spaces. Our collection is called "VADS Primitives". The plugin appears to lowercase it when exporting, but doesn't convert the spaces. It would be possible Weto update the collection names to use a hyphens, i.e. "VADS-Primitives", if that saves us extra effort.

It looks like the plugin might have some functionality to [transform it for Style Dictionary](https://github.com/lukasoppermann/design-tokens#transforming-tokens-using-amazon-style-dictionary). I have to admit I'm not too familiar with that, so perhaps we can pair up on it?

humancompanion-usds:
The plugin references this tool to get it into a format Style Dictionary can import: https://github.com/lukasoppermann/design-token-transformer

humancompanion-usds:
The plugin claims to support a description. Does Figma support a description and can I see one in the output here?

src/_data/tokens/vads.tokens.json
```diff
@@ -0,0 +1,9026 @@
...
+              ],
+              "variableId": "VariableID:571:1736",
+              "exportKey": "variables"
+            }
+          }
+        }
+      },
+      "action": {
+        "vads-color-action-focus-on-light": {
+          "description": "Focus outline color",
```
danbrady, line 5384: @humancompanion-usds Here's an example of a description on the vads-color-action-focus-on-light token.

danbrady:
> The plugin references this tool to get it into a format Style Dictionary can import: https://github.com/lukasoppermann/design-token-transformer

FWIW: Both the plugin and this tool are by the same author.

> The plugin claims to support a description. Does Figma support a description and can I see one in the output here?

Yep. We didn't add all the descriptions to the tokens yet, but we did add it to vads-color-action-focus-on-light. You should be able to see [the description here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/2224/files#diff-e192b27e30783be1ac4d166ab681ac424bbed42cf38d198d2809149b6c885414R5384). I also tagged you in an inline comment. 

(You might have to select the "load diff" link first to jump to the right spot.)

humancompanion-usds (approved):

Issue #1977, opened 2023-08-04
va-icon: Fix the sprite.svg path for the Storybook production site


jamigibbs:
## Description

The va-icon Storybook page is not loading the `sprite.svg` file but it does work in Chromatic links and locally:

- https://design.va.gov/storybook/?path=/docs/uswds-va-icon--default

![Screenshot 2023-08-04 at 12 36 26 PM](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/872479/9a1e22f8-c25b-4350-bb0e-37cbff157dbf)

![Screenshot 2023-08-04 at 12 37 01 PM](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/872479/f5511460-46ff-47a4-8721-1550aabc3c13)

The production site adds `/storybook/` to the url which we believe is what might be causing this. So we probably have to make sure the sprite file loads for both `[url]/storybook/assets/sprite.svg`  and `[url]/assets/sprite.svg` paths.

A possible solution is to add another folder to the Storybook `assets` folder so that it maps to both of these locations. The downside to this is that we would have to manage the sprite sheet in two locations for Storybook.

![Screenshot 2023-08-01 at 4 07 38 PM](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/872479/be48d13c-67cc-429c-979c-9bd1286516a2)

## Tasks
- [ ] Make sure the sprite.svg file is loading locally, in Chromatic, on Storybook prod, and also that va-icon still works in vets-website

## Acceptance Criteria
- [ ] va-icon works locally, for Chromatic, in Storybook prod, and vets-website.

caw310:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Design-System-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/1977?planning-poker) with Zenhub @Andrew565 @ataker @harshil1793 @it-harrison @jamigibbs @micahchiang @nickjg231 @powellkerry @rmessina1010

caw310:
This should be worked on before 66068 https://github.com/department-of-veterans-affairs/va.gov-team/issues/66068 as it should fix the icon issue.

powellkerry:
[...]

PR #2238, opened 2023-10-31
 Copy Storybook Assets to Root Directory using Gulp

harshil1793:
This PR introduces a new Gulp task that automates the process of copying assets from the `storybook/assets/ `directory to the root `assets/` directory. This ensures consistent asset paths across different environments, addressing the issue of assets not being found in the production environment.

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1977

### DIFF
--- a/config/gulp/storybook.js
+++ b/config/gulp/storybook.js
@@ -4,7 +4,7 @@ var task = 'storybook';
 gulp.task('move-assets', function (done) {
     console.log("Moving assets from storybook to root assets directory.");
 
-    var stream = gulp.src('./node_modules/@department-of-veterans-affairs/formation/dist/assets/*')
+    var stream = gulp.src('./node_modules/@department-of-veterans-affairs/web-components/dist/assets/*')
         .pipe(gulp.dest('assets'));
 
     return stream;
--- a/config/gulp/storybook.js
+++ b/config/gulp/storybook.js
@@ -1,10 +1,10 @@
 var gulp = require('gulp');
 
-gulp.task('move-assets', function () {
+gulp.task('storybook', function () {
     console.log("Moving assets from storybook to root assets directory.");
 
     return gulp.src('storybook/assets/**/*')
         .pipe(gulp.dest('assets/'));
 });
 
-gulp.task('default', gulp.parallel('move-assets'));
+gulp.task('default', gulp.parallel('storybook'));
--- a/config/gulp/storybook.js
+++ b/config/gulp/storybook.js
@@ -1,10 +1,13 @@
 var gulp = require('gulp');
+var task = 'storybook';
 
-gulp.task('storybook', function () {
+gulp.task('move-assets', function (done) {
     console.log("Moving assets from storybook to root assets directory.");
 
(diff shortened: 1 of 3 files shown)


config/gulp/storybook.js
```diff
@@ -0,0 +1,10 @@
+var gulp = require('gulp');
+
+gulp.task('move-assets', function () {
+    console.log("Moving assets from storybook to root assets directory.");
+
+    return gulp.src('storybook/assets/**/*')
+        .pipe(gulp.dest('assets/'));
```
jamigibbs, line 7: I think there are a few things missing from this task. 

You aren't defining the `task` name and the import `src` is not pointing to a place where the component library assets can be accessed (the sprite sheet specifically).

So you probably want to do something like this instead (you will want to confirm the destination location though to make sure it works with va-icon).

And once this runs successfully, this PR should contain the images to be checked-in so make sure it's tested locally.

```js
var gulp = require('gulp');
var task = 'storybook';

gulp.task('move-component-library-assets', function (done) {
  console.log("Moving component-library assets");
  var stream = gulp.src('./node_modules/@department-of-veterans-affairs/web-components/dist/assets/*')
    .pipe(gulp.dest('src/storybook/assets'));

  return stream;

});

gulp.task(task, gulp.parallel('move-component-library-assets'));
```

jamigibbs:
The new gulp task will not run until you import it in `gulpfile.js` before the build script runs:

Screenshot 2023-11-01 at 9 19 30 AM

jamigibbs:
@harshil1793 Can you link to the original ticket in the description of this PR so that we have a reference to why we are doing this?

config/gulp/storybook.js
harshil1793: @jamigibbs It appears that `src/storybook/assets` should contain the file already; however, as shown in the attached screenshot, the file seems to be missing from the root directory.
Screenshot 2023-11-01 at 10 53 51 AM

jamigibbs: Gotcha, so I think the important thing is that you need to [test your gulp task locally](
https://github.com/department-of-veterans-affairs/vets-design-system-documentation#to-run-locally) because the task because something does not work. When it does work, this PR should contain images to check-in for testing in prod like this.

It could be that you just need to add the import that I mentioned below for `gulpfile.js`.

gulpfile.js
```diff
@@ -7,5 +7,6 @@ require( './config/gulp/css' );
 require( './config/gulp/javascript' );
 require( './config/gulp/json' );
 require( './config/gulp/build' );
+require( './config/gulp/storybook' );
```
jamigibbs, line 10: The location of this import is important. It needs to go before `build`.

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/2238#issuecomment-1789040753

harshil1793: oh lol, I missed it. It works now.

jamigibbs (approved):
Let's give it a try! 🤞

Issue #2280, opened 2023-11-28
Font Awesome icon `fa-phone-alt` is not displaying


jamigibbs:
## Description

On design.va.gov, the phone alt icon is not displaying:

https://design.va.gov/foundation/icons#feedback

![Screenshot 2023-11-28 at 3 20 47 PM](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/872479/6f46a5ff-5253-46c0-8d83-f0657d893827)

Based on waybackmachine, it was displaying like this:

![Screenshot 2023-11-28 at 3 07 47 PM](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/872479/ed90342f-3e18-4169-9137-243323870ec7)

## Tasks
- [ ] Figure out why the icon is no longer displaying and fix it.

## Acceptance Criteria
- [ ] The phone alt icon is displaying again

jamigibbs:
From this [Slack convo](https://dsva.slack.com/archives/G01BJ3ESXL4/p1701269946156369) via Matt:

> I'm pretty sure we do use v5.15.4. The package.json just indicates the minimum version (my understanding is that is what that leading carrot means). But we could/should update it to be v5.15.4.

caw310:
Added to sprint as unplanned work.

PR #2291, opened 2023-12-04
Update Formation to latest version v9.0.1

jamigibbs:
The latest version includes an update of the Font Awesome version to use v5.15.4. This fixes the issue of the `fa-phone-alt` icon not displaying. Tested locally:

Screenshot 2023-12-04 at 9 53 13 AM

closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2280


src/assets/VCL-icon-white.svg
jamigibbs, line 1: I'm adding these generated assets just to get them into the repo but they are unrelated to Formation so I can create a separate PR if we'd like to keep them separate. 

They are getting added from a gulp script we added so that we could get some assets to display in storybook prod. See: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1977#issuecomment-1787619597

PR #2310, opened 2023-12-07
Mc v33.0.0 release

micahchiang:
## Description
- updates component library and all hard coded instances of old primary blue hex values to USWDS3
- shifts all instances of `primary-darker` to `primary-dark` and all instances of `primary-darkest` to `primary-darker`


humancompanion-usds (approved):
Thanks for doing this. Could those hex values use variables instead? I'd like to get closer to "We update token, everything magically updates across all of the places"

micahchiang:
I think eventually we could get to using variables over hex values. Challenge here is that those are declared in yml files. Need to look into how variables would work.

PR #2407, opened 2024-01-18
Updates the signature pattern with details on how to generate a PDF. [Fixes #2382]

### LINKED ISSUE #2382: Update signature pattern to include directions on how to timestamp forms
### Duplicate check

### This update is for:

Pattern

### What is the name?

Ask users for signature

- [ ] When to use this component/pattern
- [ ] Usability guidance
- [ ] Accessibility
- [ ] Implementation
- [ ] Research insights
- [ ] Package information

Today all forms are timestamped in a different way. We want to standardize how forms are marked as being generated on va.gov. We also want to capture in that timestamp/statement the IAL level of authentication that the user filling out the form had at the time. 

### Additional Context

Thoughts from Emily Theis on what should be captured:

* Enter the Veteran's signature as text, typed from the statement of truth component
  * or from their logged-in name if statement of truth component has not yet been implemented, but I think all teams should implement statement of truth componenet for sconsistency
* Enter the date in the signature field date as text
* Add a note to the bottom of the PDF indicating the form was electronically submitted via [va.gov](http://va.gov/) at datetime with X level of authentication. the date should be the date the form was submitted, not the date the PDF was generated, in case of delays or issues with submission.
  * Some forms put a random stamp at the top with just "[va.gov](http://va.gov/) submission" so we could make the placement consistent
  * From what i've heard from VBA Ideally, this note would be stamped in the "datestamp" box at the top right most PDF but that requires a little more PDF mapping per form and it appears an "at the bottom/at the top" stamp on all the pages is sufficient

humancompanion-usds:


src/_patterns/ask-users-for/signature.md
```diff
@@ -34,15 +34,64 @@ This pattern is used at the end of a form review page.
...
+>
+  <p class="vads-u-margin-y--0">
+    <strong>Generating a PDF:</strong> When generating a PDF for submission, for example when using the <a href="https://developer.va.gov/explore/api/benefits-intake">Benefits Intake API</a>, the following rules must be adhered to.
+  </p>
+</va-alert>
+
+#### Generating a PDF
+
+* **Veteran's name from Statement of truth is entered into the Signature field.** Enter the input from the Statement of truth text-input into the Requester's or Veteran's signature field on the PDF. If the form is not using the Statement of truth component yet, use the user's signed-in name. **NOTE:** All forms should implement the Statement of truth component.
+* **Current date is entered into the Date Signed field.** Enter the current date into Date Signed or Date field usually found next to the Signature field on the paper form.
```
jeana-adhoc, line 64: Might state something here about how we don't need to ask the user for the date. It can be handled on the back-end. A new team asked me this recently.

src/_patterns/ask-users-for/signature.md
```diff
@@ -34,15 +34,64 @@ This pattern is used at the end of a form review page.
...
+  uswds
+  visible="true"
+>
+  <p class="vads-u-margin-y--0">
+    <strong>Generating a PDF:</strong> When generating a PDF for submission, for example when using the <a href="https://developer.va.gov/explore/api/benefits-intake">Benefits Intake API</a>, the following rules must be adhered to.
+  </p>
+</va-alert>
+
+#### Generating a PDF
+
```
jeana-adhoc, line 62: We might need to say something here about how this is handled on the backend by engineers? I'd be nervous about designers suddenly worrying about having to design watermarks. If this guidance is followed, there's nothing for a designer to do. Could we cross link this from platform engineering documentation too?

src/_patterns/ask-users-for/signature.md
```diff
@@ -34,15 +34,64 @@ This pattern is used at the end of a form review page.
...
+### Implementation details
+
+<va-alert
+  close-btn-aria-label="Close notification"
+  full-width="false"
+  slim
+  status="warning"
+  uswds
+  visible="true"
+>
```
jeana-adhoc, line 55: I'm not used to editing dst documentation - but this alert component seems oddly placed -- it might not be though.

jeana-adhoc (commented):
Just a few comments for your review

src/_patterns/ask-users-for/signature.md
humancompanion-usds: Added a note to explain that. Good catch.

src/_patterns/ask-users-for/signature.md
humancompanion-usds: I added a note that this is specifically for engineers. Yes, the intention is to link to this page from the Platform engineering documentation.

src/_patterns/ask-users-for/signature.md
humancompanion-usds: I want folks to not miss it. I don't use alerts very often in guidance. But, Alert - Slim can be used beneath a header like it is here. I'm careful not to violate the Design System's own rules in the Design System. :-)

Andrew565 (approved):

Issue #2301, opened 2023-12-06
Switch Sketch links to Figma for all components


humancompanion-usds:
### Duplicate check

- [X] I've searched for any related issues and avoided creating a duplicate issue.

### This update is for:

Component

### What is the name?

All components

### What is the nature of this update?

- [ ] How to build this component/pattern
- [ ] When to use this component/pattern
- [ ] When to use something else
- [ ] Usability guidance
- [ ] Accessibility
- [ ] Implementation
- [ ] Research insights
- [ ] Package information
- [ ] Addition to Forms Library documentation
- [ ] Update to existing Forms Library documentation

### What problem does this solve?

All links to Sketch will now switch to Figma

humancompanion-usds:
I need the Example library for this.

humancompanion-usds:
@caw310 and @danbrady - Let's discuss in our chat tomorrow how we are progressing with the Example library and how I can tell which components are safe to link to. Perhaps we bring back the dots for this one?

danbrady:
@humancompanion-usds I went through all the Sketch links in each component on the doc site to see where they are pointing to. About half link to the Example Library while the other half link to Pattern Library (and a few have no links, link to Figma already, or have an error). Both libraries have a lot of overlap with each other, however, oftentimes an example might be in one but not the other. What's the best way to sort through this so we don't repeat mistakes or cause confusion? I can think of a couple different options:

1. Follow one of the libraries. Which ever is newer or most recently maintained? (Both have recent Last Modified dates so I'm not sure which one it would be.)
2. Combine examples from both (The "union" of the two.) But I'm not sure if there's old/outdated examples in there.
3. Follow whatever library the component links to on the docs site. 
4. Other ideas...?


Below is the list of what components are currently linked to:


#### Links to Example Library
* Breadcrumbs
* Button
 * Button
 * Button Pair
* Divider
* Form
  * Checkbox
  * Memorable Date
  * Number Input
  * Privacy Agreement
  * Progress Bar - Segmented
  * Radio Button
  * Select
  * Text Input
  * Textarea
* Header
 * Header
 * Minimal
* Modal
* Pagination
* Process List



#### Links to Pattern Library
* Address Block
* Alert - Expandible
* Back to Top
* Banner 
  * Maintenance
  * Official Gov
  * Promo
* Card
* Featured Content
* Form
  * Date Input
  * File Input 
  * Need help?
  * Penalty Notice
  * Prefill
[...]

PR #2409, opened 2024-01-19
Adding first pass of examples

danbrady:
Applies to #2301.


humancompanion-usds (changes requested):
This changed a lot more than you intended. I suspect you have a linter or something that is auto-changing the * to - for the bullet items. My editor does the reverse. I'm not sure which is right but I've standardized on *. I hate to ask you to do this but could you revert those changes?

danbrady:
You're right. I added a local workspace .prettierignore and custom markdown linter to not do that again. Please re-review. Thanks!

humancompanion-usds (approved):

Issue #2423, opened 2024-01-24
Update documentation site colors to USWDS v3


powellkerry:
## Description
- Update the colors on the documentation site to be the USWDS v3 colors
- If possible configure the site to automatically pull these colors from the css-library

## Details
Pages needing updates:
[Background color](https://design.va.gov/foundation/utilities/background-color)
[Border color](https://design.va.gov/foundation/utilities/border#border-color)
[Text color](https://design.va.gov/foundation/utilities/text-color)

yaml file containing colors: _data/colors.yml

## Tasks
- [ ] Update colors to v3
- [ ] Configure site to use css-library variables/tokens

## Acceptance Criteria
- [ ]

PR #2444, opened 2024-01-30
Update to tokenize colors and update them to v3

powellkerry:
- Update colors.yml to include v3 tokens defined in css-library
- Update templates to display tokens

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2423


humancompanion-usds (approved):

src/_foundation/color-palette.html
```diff
@@ -65,15 +65,15 @@ <h2 id="primitive-color-tokens">Primitive color tokens</h2>
 </div>
 
 
-<h2 id="v1-colors">v1 colors</h2>
+<h2 id="v1-colors">v3 colors</h2>
```
harshil1793, line 68: Not a blocker but it would be great if id matches the heading content.

Issue #2391, opened 2024-01-10
Update the guidance template to include mobile


caw310:
## Description
Update the component guidance page template on design.va.gov to include guidance and usage for mobile components

## Details
Review examples of mobile component guidance from Jessica and use those examples to update our templates and our guidance for a few components.

## Tasks
1. [ ] Bring the following components into VADS:
  * [x] Button
  * [x] Link 
  * [x] Alert
2. [ ] Update component template to have spots for mobile
3. [ ] Create a mobile-only component template for components that don't yet have a web equivalent

## Acceptance Criteria
- [ ] Component guidance page has section for mobile

humancompanion-usds:
@caw310 - I'd like us to create a new Epic for integrating the flagship mobile app with the VA Design System. This is more than Maintenance. Also, I'd really like maintenance to get phased out as an epic. 52 issues is too many for one epic. It's a dumping ground. Thanks.

caw310:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Design-System-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/2391?planning-poker) with Zenhub @danbrady @LWWright7

humancompanion-usds:
From @jessicawoodin:  Here are 3 examples of components that have shared documentation with VADS. For each component, I've linked to the documentation on the mobile doc site and the Google doc. In the Google doc, I copied in the VADS documentation and used track changes to see where mobile might be different.

* Button ([Mobile doc site](https://department-of-veterans-affairs.github.io/va-mobile-app/design/Components/Buttons%20and%20links/Button) | [Google doc](https://docs.google.com/document/d/1oZgsJytSN3GNVAPN9wzfOz3WOeoMuNG0CWDTg2BRehs/edit#heading=h.vj9zwf7j22yt))
* Link ([Mobile doc site](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/Flagship%20design%20library/Components/Buttons%20and%20Links/ClickForActionLink) | [Google doc](https://docs.google.com/document/d/1ak2hC3ceOla5hMG9sjjCRloup2Ee-D0D_grsWbJMfwk/edit#heading=h.vj9zwf7j22yt))
[...]

PR #2514, opened 2024-02-20
Starts to bring mobile buttons into VADS. [Amends #2391]

humancompanion-usds:


humancompanion-usds:
Started a new, more recent, branch for this work.

PR #2551, opened 2024-02-29
Moving v3 examples to be above v1

danbrady:


humancompanion-usds (changes requested):
This is a good start but it's not just changing the order. We also want to swap the variations to all be v3, not just the default, and then make v1 just show the default option (as we had done for v1).

danbrady:
Oh man, I didn't realize that. My mistake. I was thinking "This is a simple copy/paste!". Now you know why I combined the two tickets! I didn't realize I'll need to update/insert stories. It probably makes sense to split the issues back apart.

PR #2575, opened 2024-03-11
WCAG 2.2 updates

briandeconinck:
Minor updates to places that reference WCAG 2.1


briandeconinck:
Noting here that this is part of work happening in department-of-veterans-affairs/va.gov-team/issues/49681

it-harrison (approved):

Issue #2197, opened 2023-10-12
Icon web componet - Documentation


caw310:
## Description
Add or update documentation and guidance for Icon on design.va.gov. When adding a new component, use the [component documentation template](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/src/_components/template.md).

## Details
https://design.va.gov/foundation/icons

Be sure to use the appropriate [maturity scale](https://design.va.gov/about/maturity-scale) option. If this is a new component that has not yet gone through Staging Review, use the "Use with Caution: Candidate" label.

## Tasks
- [ ] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [ ] Request documentation draft review
- [ ] Once documentation is approved, publish to design.va.gov

## Acceptance Criteria
- [ ] Component documentation is published on design.va.gov

caw310:
Swapped this out with one thing per page so moving to next sprint.

danbrady:
Most of the icons should show once @Andrew565 adds the icons to component-library and brings it into the doc site. However, there are a few minor edits needed: There is no "pill" icon, so that's swapped with a medicine bottle ("medication"). Also we have two new icons for the Apple App Store and Google Play. I will work with Andrew to update the icon table in the documentation and will submit a PR.

humancompanion-usds:
@danbrady - Any remaining work to be done here?

PR #2637, opened 2024-03-27
Adds the va-icon component to guidance and updates the icons on the foundational page. [Amendsd #2197]

humancompanion-usds:


Andrew565 (approved):

danbrady (commented):
@humancompanion-usds I see there are a few missing USWDS icons here. Some of the icon string names may be different. I think Lilly might have those names in the Mural analysis. 

Also, what do you think about updating the table to have the v3 icon names? It might make it easier for developers to find their v3 icon equivalents. Something like this:

image

Would you want me to submit a separate PR with this update and corrected icon names after this publishes?

jamigibbs:
> Also, what do you think about updating the table to have the v3 icon names? It might make it easier for developers to find their v3 icon equivalents. Something like this:

Oh this is nice. If this happens, I can also update the ESLint rule to point to this table for the `icon` (name) prop.

PR #2749, opened 2024-04-21
Update links.md

megzehn:
updating guidance to remove sub-domains and guidance that external links open in a new tab


megzehn:
HI @mnorthuis and @DanielleThierryUSDSVA, I made some minor edits to this page (on the content side) to remove the VA subdomain links that got added in there earlier. Could you review when you get a chance? Thank you so much!

mnorthuis:
Since the content mentions what to do when we open a new tab, it seem like we should make reference to the fact that that is not the norm, and direct them to the [link component](https://design.va.gov/components/link/) for interaction and behavior guidance.

Did we open a ticket for the DST to fix that guidance as well?

humancompanion-usds:
@mnorthuis and @DanielleThierryUSDSVA - Is there an issue to remove the subdomain exceptions from code? Has that already been done? 

@megzehn - Can you connect this PR to the appropriate issue? I just want to be sure we save the history trail here because we've gone in circles on this question now a few times.

mnorthuis:
@humancompanion-usds I have not created an issue, but definitely can if you let me know who it needs to go to.

humancompanion-usds:
@mnorthuis - I'm going to guess sitewide would be able to handle it, or redirect it should it not fall under their purview.

humancompanion-usds (approved):

PR #2789, opened 2024-05-01
Add missing icons and requesting new icons guidance

danbrady:


src/_foundation/icons.html
```diff
@@ -66,6 +70,33 @@ <h2 id="{{ section[0] | downcase }}">{{ section[0] }}</h2>
...
 {% endfor %}
 
+<h2 id="requesting-a-new-icon">Requesting a new icon</h2>
+<p>
+  If your team needs a new icon and wants to suggest adding it to the design system, follow these steps:
+</p>
+
+<ol>
+  <li>
+    <strong>Check Existing Icons</strong>:</strong>
```
Andrew565, line 80: There's an extra </strong> closing tag here. Not hurting anything, just pointing it out.

src/_foundation/icons.html
```diff
@@ -66,6 +70,33 @@ <h2 id="{{ section[0] | downcase }}">{{ section[0] }}</h2>
...
+
+<ol>
+  <li>
+    <strong>Check Existing Icons</strong>:
+    Look through the current icon set to ensure a similar icon doesn't already exist. We aim to maintain consistency in
+    semantic use by avoiding duplicating similar icons.
+  </li>
+
+  <li>
+    <strong>Explore Material Icons</strong>:
```
humancompanion-usds, line 86: I'd want them to go to USWDS first and foremost, then to Material if they cannot find what they need. https://designsystem.digital.gov/components/icon/

humancompanion-usds (changes requested):
Need to point them to USWDS first, then to Google. I would down play calling it Material. Folks outside of the Platform may get the notion that we are using Material, which we're really not.

danbrady:
Thanks, @humancompanion-usds. I updated the content to steer folks to USWDS before Material. Let me know if that makes more sense. Thanks!

humancompanion-usds (approved):

PR #2823, opened 2024-05-15
Adds button - icon guidance. [Fixes #1559]

### LINKED ISSUE #1559: Experimental Design - Button - Icon
## What
I'd like to add a Button variant named Icon thus Button - Icon. It's a tertiary button style that is lightweight visually and can be used alongside Link variations.

## Purpose
Our Button - Primary and Secondary variations are visually dominant and large. When we have a number of secondary buttons clustered together we lose visual hierarchy and we hypothesize that users take longer to find the action they'd like to take. This Button style looks to repair those issues.

## Usage
* Button - Icon could be used inside of a Card, Modal, or other container where there may be multiple actions in close proximity. 
* Button - Icon could be used in Secure Messaging and other MHV applications were multiple actions of equal prominence are in close proximity and requite buttons.

## Behavior
These buttons start with an icon and then uppercase text. Upon hover state they have a background and the icon & text are a darker color.

## Examples
New Button - Icon instances@2x

Solution E 2 - Close is a Button - Icon@2x

## Accessibility
These mocks were reviewed by Brian, Josh, and Martha. This gives teams flexibility when we want to have material honesty in using links and buttons correctly.

## Guidance
These Button - Icon will only be available initially to be used inside of a container. Today containers are Card, Modal, and application specific boxes with a darker background for grouping similar actions. They are not to be used in place of primary or secondary buttons outside of these containers. 

## Research (optional)
No research has been conducted as of yet. The notion is to make these available to teams who have been asking for solutions in this space.

## Next steps
Plan is to present this component on 2/17.

humancompanion-usds:


src/_components/button/button-icon.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Usage
+
+### When to use Button - Icon
+
+* **In a Card or other container alongside a primary link.** This component is mainly intended to sit alongside a primary link, typically a [Link - Active]({{ site.baseurl }}/components/link#active) variation, within a [Card]({{ site.baseurl }}/components/card). This allows the link to visually have the affordance as the primary action and the Button - Icon the affordance of a secondary action.
+
+### When to consider something else
+
+* **When a secondary button will suffice.** Button - Icon does not replace the [Button - Secondary]({{ site.baseurl }}/components/button#default---secondary). Only use Button -Icon when Button - Secondary would visually dwarf a link primary action.
```
danbrady, line 44:
```suggestion
* **When a secondary button will suffice.** Button - Icon does not replace the [Button - Secondary]({{ site.baseurl }}/components/button#default---secondary). Only use Button - Icon when Button - Secondary would visually dwarf a link primary action.
```
Added space

src/_components/button/index.md
```diff
@@ -70,12 +71,12 @@ anchors:
 
 * **Avoid using many primary buttons on a single page or section.** Pages with many primary buttons reduces their impact and make it harder for users to know what to do next.
 * **Arrows are reserved.** Arrow icons should only appear for "Back" and "Continue" buttons that appear in forms.
+* **Use icons only when necessary.** [Icons]({{ site.baseurl }}/foundation/icons) can be used in buttons when additional clarity is required and the icon is highly relevant to the action. Icons should not be used for decoration.
```
danbrady, line 74: Just a clarification - Are we saying that icons can go in regular buttons here or that they should use Icon - Button?

Also, currently, in both Button and Button - Icon there is not a way to use *any* icon in them. Essentially, we've hardcoded a few variations with icons. Do we want to mention the process on how to go about suggesting a new variation with icon?

src/_includes/content/button-labels.md
```diff
@@ -1,13 +1,12 @@
...
-- Use sentence case for button labels.
-- Make button labels as short as possible.
-- Use “trigger words” that people recognize and clearly signal what will happen when they click the button.
-- Keep the character limit for button labels to 35 characters. Button labels should be as short as possible with “trigger words” that your users will recognize to clearly explain what will happen when the button is clicked (for example, "Back" or "Continue"). 
+For simple navigation to lead people between pages, use links instead of buttons.
 
+* Use sentence case for button labels (except for Button - Icon).
+* Make button labels as short as possible.
+* Use "trigger words" that people recognize and clearly signal what will happen when they click the button.
+* Keep the character limit for button labels to 35 characters. Button labels should be as short as possible with "trigger words" that your users will recognize to clearly explain what will happen when the button is clicked (for example, "Back" or "Continue").
```
danbrady, line 10: We use "trigger words" on the Button Labels content page as well so that's likely why it carried over here, but I wonder if "action words" is more appropriate?

Also, should we link to that Button labels content guidance page here?

src/_components/button/index.md
humancompanion-usds: We're saying regular Buttons can have icons but only when necessary, which is rare. We say this a lot in practice but don't have it written down. Agree that it isn't possible in the web-components but there are lots of buttons that are not yet va-button instances. That's where I see icons in buttons. 

Yes, to talking about suggesting a new icon. I'll point to your doc in foundations.

src/_includes/content/button-labels.md
humancompanion-usds: This is the Button labels guidance. It's an include that shows up in both places. Agree about "trigger words". Will change.

danbrady (approved):

babsdenney (approved):

PR #2867, opened 2024-05-23
bump component-library to 42.1.0

it-harrison:
This PR bumps the version of component-library to 42.1.0

**After updating formation and rem values**:

Screenshot 2024-05-29 at 2 04 06 PM

Screenshot 2024-05-29 at 2 07 07 PM

Screenshot 2024-05-29 at 2 08 22 PM


jamigibbs (commented):
This will probably also need an update to Formation and also sweep the repo files to convert `rem` values to the 16px base font size. Make sure to verify this update locally before merging the PR. 🤞🏼

jamigibbs (approved):

jamigibbs (commented):
@it-harrison Is this site already on 16px root font? I might not have checked that when I made my comment about updating the rem. :/ 

If it was already at the root font size, do we even need to change those values?

Screenshot 2024-05-29 at 2 40 44 PM

jamigibbs (approved):

PR #2892, opened 2024-05-30
Update index.md

coforma-jamie:
Requesting that we add CAIA's slack handles to the CAIA section

### DIFF
--- a/src/_about/accessibility/index.md
+++ b/src/_about/accessibility/index.md
@@ -84,7 +84,12 @@ It is the intention of this team to improve our testing standards to more accura
 
 The Sitewide Content, Accessibility, and Information Architecture (CAIA) team can help you meet VA.gov’s accessibility standards. Whether you’re starting a new product or refining an existing one, contact the CAIA team as early as possible to create accurate, consistent, accessible, and equitable digital services for Veterans.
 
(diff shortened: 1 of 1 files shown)


humancompanion-usds (changes requested):
Hmm, this is venturing into territory I would prefer to avoid. One rubric we've been standardizing on is that implementation process details stay in the Platform Content site rather than come into the Design System. Thus you'll note we mention the Collab Cycle, but we link to the Collab Cycle section of platform content rather than including details on how to engage with that process here. Linking to Platform Content or GitHub is fine but we shouldn't bring that information into the Design System. Thus if you want to make a page in Platform Content that talks about how to reach CAIA then you could link to that page from here. Perhaps linking to the "Content, Accessibility, and Information Architecture support - Sitewide CAIA" page (https://depo-platform-documentation.scrollhelp.site/collaboration-cycle/sitewide-content-and-ia-intake-request) from here would work?

coforma-jamie:
@humancompanion-usds I think that's a great idea. Is there a way to edit this PR, or do we need to make a new one?

humancompanion-usds:
@coforma-jamie - That's up to you. You can continue to commit changes on this branch. If it's easier to start over just say so and I can close this.

humancompanion-usds (approved):

PR #2937, opened 2024-06-13
Clarify single-page L&L guidance use

danbrady:
FYI @jeana-adhoc @allison0034


Andrew565 (approved):

humancompanion-usds (approved):
Fine but we should (someday) fix up the page layout so we don't need that p tag.

Issue #2415, opened 2024-01-22
Update content style guide for Title tags


megzehn:
## What
When forms use the minimum header design, the question for each page in the form (like "Are you a Veteran?") becomes the h1. Right now, our guidance suggests that we pull this in for the title tag (like "Are you a Veteran? | Veterans Affairs") A11y, IA, and content folks propose that for forms (or anywhere we use the minimum header design), we use a title tag that follows the pattern of [title] | Veterans Affairs instead of the current guidance of [h1] | Veterans Affairs.

## Purpose
If a general form question is in the title tag, it's less meaningful. I.e.: "Are You a Veteran? | Veterans Affairs" vs "File For Disability | Veterans Affairs"

## Usage
Wherever the h1 is not a descriptive name for the tool or form, we should consider using the title for the title tag instead of the h1. 

## Accessibility
If someone is filling out 2 or 3 forms and each form is in its own window, the person may be alt + tabbing between them. If a screen reader is calling the names out, it would say "[Question] | Veterans Affairs” — and the person wouldn’t be able to tell the forms apart.

## Guidance
The title tag should use the same title as the form name throughout the full form. 

## Code (optional)
> Include any existing code.

## Next steps
Asynchronous approval would be great. But we can attend if there are any questions or suggestions for how to approach this issue. @jennymayoco and @kristinoletmuskat have been involved in this discussion, so they can help answer any follow-up questions. 

-----

Submit requests to join an upcoming Design System Council meeting in #platform-design-system.  

During the meeting, the Design System Council Working Group will evaluate the request and make a decision. 

If your request is approved, you can [add your component or pattern to the system](https://design.va.gov/about/contributing-to-the-design-system/add-a-component-or-pattern-once-approved). If you have any questions on how to add your component or pattern to the system, please reach out to the Design System Team at #platform-design-system.

humancompanion-usds:
Carol and I discussed this one this morning. We think this is just a guidance clarification. I don't think anyone intended for the title to remain the question in the H1. Or at least I don't think anyone is advocating for that. The Forms team is working on templatizing One thing per Page patterns which I'll put in the design system. When I do that I'll also add this guidance. But I would appreciate a draft of the guidance, which is mostly here already, if you have the time to do that.

megzehn:
Great! Thanks, Matt! @jennymayoco are you able to provide a draft for @humancompanion-usds?

kristinoletmuskat:
[...]

PR #2986, opened 2024-07-02
Update formatting of title tags examples

erinrwhite:
Updated formatting on the examples in the [Exception 2: forms and subtasks title tags section](https://design.va.gov/content-style-guide/title-tags#exception-2--forms-and-subtasks) to be more clear. The current formatting doesn't exactly match the text in the word doc shared in #2415, which is resulting in incorrect guidance. I broke the examples out into a nested UL for easier understanding.


danbrady (approved):

erinrwhite:
thanks @danbrady! I tagged @humancompanion-usds but I'm not sure if he needs to review this change. Leaving to y'all.

Issue #2655, opened 2024-04-03
Update File Input - Documentation


caw310:
## Description
Update documentation and guidance for File Input on design.va.gov to include guidance for uploading multiple files. 
When adding a new component, use the [component documentation template](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/src/_components/template.md).

## Details
[add links to any design documents]

[add link to Figma component page]

Be sure to use the appropriate [maturity scale](https://design.va.gov/about/maturity-scale) option. If this is a new component that has not yet gone through Staging Review, use the "Use with Caution: Candidate" label.

## Tasks
- [ ] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [ ] Request documentation draft review
- [ ] Once documentation is approved, publish to design.va.gov

## Acceptance Criteria
- [ ] Component documentation is published on design.va.gov

caw310:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Design-System-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/2655?planning-poker) with Zenhub @babsdenney @danbrady @LWWright7

babsdenney:
[Chromatic link for file input examples](https://2642-file-input-update--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/docs/uswds-va-file-input--default)

caw310:
Will most likely carry over as documentation is pending Storybook update.

humancompanion-usds:
I believe we've done all we can do at this point until we convince a team to deploy using this new component. Then we can update the Ask a user for files pattern. But the component work seems done to me. Reopen is you disagree.

PR #3041, opened 2024-07-17
Corrects paths in file input component doc. Corrects 2 examples in radio button. Adds some file pattern images for future use. [Amends #2655]

humancompanion-usds:


babsdenney (approved):
Thank you!

Issue #2784, opened 2024-04-26
Ask a user for their... Housing status/situation


jeana-adhoc:
### Duplicate check

- [ ] I've searched for any related issues and avoided creating a duplicate issue.
**NOTE: I missed the related issue and this is now a duplicate ticket.**
  - https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1371

### This update is for:

Pattern

### What is the name?

Ask a user for their... Housing status/situation

### What is the nature of this update?

- [ ] How to build this component/pattern
- [X] When to use this component/pattern
- [ ] When to use something else
- [X] Usability guidance
- [ ] Accessibility considerations
- [X] Content considerations
- [ ] Implementation
- [ ] Package information
- [ ] Addition to Forms Library documentation
- [ ] Update to existing Forms Library documentation

### What problem does this solve?

Provides a plain language way to ask Veterans their housing status that allows them to answer the question accurately and without shame

### Additional Context

We worked with @DanielleThierryUSDSVA  on this content

[I pulled out the specific pieces into this figma file](https://www.figma.com/file/PgZ1U35raBde6s3viqotpA/WIP-DSC---Pattern---Ask-Users-for-Housing-Situation?type=design&node-id=1-8&mode=design&t=nuRmgBaAuoNH1Vmz-11). 

You can see it in action in this form - [Request priority processing for an existing claim 20-10207](https://staging.va.gov/supporting-forms-for-claims/request-priority-processing-form-20-10207)

[You can also see it in use in the forms figma file here](https://www.figma.com/file/DA6OjFz20kUa9ecElv9zPE/WIP---20-10207-Priority-Processing?type=design&node-id=0-1&mode=design&t=hzwRSIpijLvb77Ia-11)

<img width="566" alt="0-Homeless-Q-1" src="https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/117098918/c18e4269-96aa-4cf8-92f3-397571b2ca5b">
<img width="509" alt="1-Homeless-None" src="https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/117098918/a581ea8a-f646-4a26-9551-f05e52496e8b">
<img width="545" alt="1 1-mailing-address" src="https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/117098918/c294634c-4be6-449d-a872-5e07b48a54d9">
[...]

PR #3043, opened 2024-07-17
Start of housing status pattern. [Amends #2784]

humancompanion-usds:


jeana-adhoc (approved):
No changes! Looks good to me.

danbrady (approved):

Issue #3040, opened 2024-07-17
Update documentation for ARIA label on telephone numbers


rsmithadhoc:
## Description

Update the guidance on the [design system documentation for the telephone number label](https://design.va.gov/content-style-guide/dates-and-numbers#links-and-aria-labels) based on the changes and new guidance outlined in https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2884

## Tasks
- [ ] Update guidance on the site to remove the notes about ARIA labels for telephone.
- [ ] Get it reviewed by CAIA, they maintain the Content sections.

## Acceptance Criteria
- [ ] Design system documentation has been updated to remove the telephone label notes.

megzehn:
Hi @rsmithadhoc! Could you confirm whether or not we want to keep the a href tags that we currently have on phone numbers?

For example, if we have this coding:

`<a href="tel:2021231234" aria-label="2 0 2. 1 2 3. 1 2 3 4."> 202-123-1234 </a> (<a href="tel:711" aria-label="TTY. 7 1 1.">TTY: 711</a>)`

under the new guidance, should it be this?

`<a href="tel:2021231234"> 202-123-1234 </a> (<a href="tel:711">TTY: 711</a>)`

We're starting to roll out this guidance in production on our Drupal pages. There are many places where we've added aria labels to phone numbers, so I want to make sure we're making the right changes. 

CC @RLHecht @saratorres2 @laurwill

megzehn:
Update: @artsymartha68 confirmed we're keeping the `<a href>` tags on phone numbers and TTY: 711. We're just removing the `<aria-label>`

artsymartha68:
Correct, @megzehn , thanks for verifying!

Issue #2884, opened 2024-05-29
Remove aria-label from va-telephone component


rsmithadhoc:
## Description

Based on a discussion with the accessibility group at VA.gov, we decided it would be better to remove the `aria-label` containing the spaced out phone number (e.g. `aria-label="8 7 7. 3 4 5. 9 8 7 6."`). This was intended to help screen reader users by reading each individual number, but hurts the experience for speech recognition users. Because screen reader users have the option to read out numbers individually, the prevailing guidance around this has changed.

## Details

Explanation given by Siteimprove, provided by Martha:

> 
> I assume that you put spaces in the aria-label so that the phone number would be announced as single digits. I agree that there is a benefit for screen reader users in having the phone number announced as single digits. The problem is that this benefit for screen reader users makes the experience worse for speech recognition users, and in our opinion, the latter outweighs the former. So this is not a false positive. It's a true positive.
> This is a difficult trade-off. If there was a way to change the web page to give both screen reader users and speech recognition users the ideal experience, then I would recommend it. But there is no known way to do that - not by changing the web page. Fortunately there are ways to do it by changing the screen reader configuration. I've written more about that below.
> This case regarding spaces in phone numbers is not straightforward. Many people in the accessibility community have discussed it for years. That this is a true positive isn't just Siteimprove's opinion - rather, it's informed by the ACT (Accessibility Conformance Testing) group. The most recent best practice is mentioned in this post https://github.com/act-rules/act-rules.github.io/issues/1615#issuecomment-1525874462. The main points of that post are:
> - Regarding the argument of giving an aria-label with spaces (eg. aria-label="4 1 6. 5 5 5. 0 1 2 3."), so that the screen reader will announce the numbers one by one (eg. "four, one, six"): we now see that argument as misguided. It's misguided because screen readers can be configured by their users to announce the numbers one by one. So for the web page author to add spaces like that is unnecessary.
> - If the web page author adds spaces like that, they are taking choice away for the screen reader user. (Because what if a screen reader user wants it to be read as big numbers eg. "four-hundred-and-sixteen"?) So for the web page author to add spaces like that is undesirable.
>
[...]

PR #3124, opened 2024-08-02
Remove ARIA label guidance for phone numbers

rsmithadhoc:
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3040

This removal is based on the background outlined in this issue https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2884 and discussions with the a11y group at VA. We have removed the ARIA label from the telephone component itself, so this guidance should also be updated to remove that guidance for links, as spacing out the phone number in a label is no longer needed.


rsmithadhoc:
Hi @DanielleThierryUSDSVA and @bethpotts,

Can either of you give this a review on behalf of CAIA, since you maintain this section?

Thanks!

humancompanion-usds:
@megzehn @laurwill just an FYI that the aria labels are coming out of the content style guide page on phone numbers.

megzehn:
Thanks for flagging this for us! 

I posted a question in #3040

PR #3183, opened 2024-08-14
Updates guidance on race and ethnicity.

humancompanion-usds:


src/_patterns/ask-users-for/race-ethnicity-or-origin.md
```diff
@@ -13,19 +13,38 @@ anchors:
...
 
 ## How to design and build 
 
 ### How this pattern works
 
-* **Give users the option of picking more than one ethnic group.** Always use checkboxes so that users can identify with multiple ethnicities. 
-* **Provide a way to opt-out of answering.** A checkbox labeled "Prefer not to answer" should be provided.
\ No newline at end of file
+* **Ethnicity is currently a distinct question that precedes race and is limited to one response.** Ethnicity is currently separate (this may change in future) and must only accept one response.
+* **Give users the option of picking more than one race or origin.** Always use checkboxes so that users can identify with multiple races.
```
danbrady, line 48: Do we need to specify how the race/origin checkboxes validation should behave if the user chooses **Prefer not to answer**? For example, if **Prefer not to answer** is selected, all other checkboxes should be unchecked and/or disabled. Also the inverse applies, **Prefer not to answer** should be unchecked if any other option is selected.

Andrew565 (approved):

src/_patterns/ask-users-for/race-ethnicity-or-origin.md
humancompanion-usds: Yes, good call. I will add.

PR #3193, opened 2024-08-16
Remove outdated links from homepage and footer; update image for developer docs link on homepage

erinrwhite:
1. Removed these links:

  - Removed [VA Playbook](http://www.va.gov/playbook/) link from site footer. Playbook link appears to have been retired and redirects back to VA.gov.
  - Removed the [VA Digital Service Handbook](https://department-of-veterans-affairs.github.io/va-digital-service-handbook/) link from design.va.gov homepage. This site has not been updated in several years and the repo in GitHub has been archived. 
    - **Note:** Removing this link results in the "Resources" section having three, rather than four, sections. Screenshot below.
2. Added image for "developer documentation" link in resources section of Design System homepage; was a GitHub logo but now uses the same image that is used on the Platform Website.

**Preview of "Resources" section:**

Screenshot 2024-08-16 at 4 15 57 PM


humancompanion-usds:
Who asked for the links to the handbook to be removed and what has replaced it?

humancompanion-usds:
@erinrwhite - No worries. I'm just reluctant because I know the folks who built the handbook and, at one time, it was very important. While it it true that no one owns it, much of what is there is still true to how we operate. I'm a little reluctant to lose that history. It's probably right to remove links to it from here but, without something to replace it that talks about how OCTO operates there is a bit of a vacuum. Let me think on this.

humancompanion-usds (approved):
Fine. I acquiesce. Long live the Playbook and all of the other resources published by Civic Tech orgs that are now RIP.

PR #3255, opened 2024-09-03
Add info banner about known issues

danbrady:


jamigibbs (approved):
Good idea to add this

Issue #3240, opened 2024-08-28
Update $small-desktop-screen  breakpoint - Documentation


caw310:
## Description
Add or update documentation and guidance for $small-desktop-screen breakpoint to 1024px on design.va.gov. 

## Details
Once the development work is done, the guidance needs to be updated. 
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3233

## Tasks
- [ ] Write or update component documentation for design.va.gov
- [ ] Coordinate with accessibility specialist to create or update accessibility guidance
- [ ] Request documentation draft review
- [ ] Once documentation is approved, publish to design.va.gov

## Acceptance Criteria
- [ ] Component documentation is published on design.va.gov

```[tasklist]
### Tasks
- [ ] https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3233
```

caw310:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Design-System-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/3240?planning-poker) with Zenhub @babsdenney @danbrady @LWWright7

danbrady:
This is blocked by #3233

PR #3267, opened 2024-09-10
Update breakpoint table and docs page with info about new breakpoint names

jamigibbs:
I'm taking the new names and values from the table in [this ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3217):

| Current VADS Name | USWDS v3 Name |
| --------------------------- | ------------------------- |
| xsmall-screen (320px)          | mobile (320px) |
| small-screen (480px) | mobile-lg (480px) |
| small-desktop-screen (1008px) | desktop (1024px) 
| large-screen (1200px) | desktop-lg (1200px) |

**Old**

Screenshot 2024-09-10 at 2 19 23 PM

**New**
Screenshot 2024-09-10 at 3 51 18 PM

closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3240

### DIFF
--- a/src/_includes/_breakpoint-names.html
+++ b/src/_includes/_breakpoint-names.html
@@ -2,7 +2,6 @@
   <thead>
     <tr>
       <th>VADS Breakpoint Sass Variable</th>
-      <th>Width</th>
       <th>USWDS Breakpoint Custom Property</th>
       <th>Width</th>
     </tr>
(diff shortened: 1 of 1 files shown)


src/_foundation/breakpoints.md
```diff
@@ -17,6 +17,10 @@ anchors:
 
 ## Names and values
 
+<va-alert status="info" slim>
+  <p class="vads-u-margin-y--0">The VA Design System is currently transitioning to breakpoint names that more closely align with USWDS. The table below lists the original VADS breakpoint names and its new corresponding name.</p>
+</va-alert>
```
jamigibbs, line 22: I thought we might want to callout the transition period here. Is this what we want to say?

Screenshot 2024-09-10 at 2 33 50 PM

src/_foundation/breakpoints.md
```diff
@@ -17,6 +17,10 @@ anchors:
...
+
 There are six key breakpoints where the design of any page layout, utility, or component may change.
```
jamigibbs, line 24: > six key breakpoints

The original table showed five breakpoints (??) and now it shows 10 breakpoints so maybe this should just be removed?

src/_includes/_breakpoint-names.html
```diff
@@ -1,30 +1,42 @@
...
       <td><code class="site-utility-class vads-u-display--inline-block vads-u-font-size--base">$small-screen</code></td>
       <td>481px</td>
+      <td><code class="site-utility-class vads-u-display--inline-block vads-u-font-size--base">$mobile-lg</code></td>
+      <td>481px</td>
     </tr>
     <tr>
       <td><code class="site-utility-class vads-u-display--inline-block vads-u-font-size--base">$medium-screen</code></td>
       <td>768px</td>
+      <td><code class="site-utility-class vads-u-display--inline-block vads-u-font-size--base">$medium-screen</code></td>
+      <td>768px</td>
```
jamigibbs, line 27: `medium-screen` is technically _not_ a USWDS breakpoint but we're listing it under that column because it'll still be used. Maybe we want to table header named something else? Like just "New VADS Breakpoint sass variable"?

src/_foundation/breakpoints.md
micahchiang: If we're showcasing sass variables here, I think we probably want to provide some guidance around how teams can import them. I don't think we're currently importing the scss tokens file into vets-website, so teams may not know they can't use them without additional configuration. 

Alternatively, we could replace the `USWDS breakpoint sass variable` column with css custom property names instead, since those are already readily available, and we may want to use them anyway. 

We should also add an entry for `tablet (640px)`, since that's available for use as well

jamigibbs: > If we're showcasing sass variables here, I think we probably want to provide some guidance around how teams can import them. I don't think we're currently importing the scss tokens file into vets-website, so teams may not know they can't use them without additional configuration.

> Alternatively, we could replace the USWDS breakpoint sass variable column with css custom property names instead, since those are already readily available, and we may want to use them anyway.

@micahchiang That's a really good point. I would very much prefer using the custom properties so I personally lean towards updating the column. I'll do that and then we can have a look to see if it makes sense alongside the sass stuff.

> We should also add an entry for tablet (640px), since that's available for use as well

Ah! This explains why it says there are "six breakpoints" but I only counted five! :) I'll get it added.

jamigibbs: @micahchiang Here is the updated table:

Screenshot 2024-09-10 at 3 15 51 PM

src/_foundation/breakpoints.md
danbrady: Yeah, taking out the number of breakpoints makes the sentence kind of moot, imo. So I think the whole sentence can be removed.

src/_foundation/breakpoints.md
jamigibbs: Based on @danbrady's idea to have [one width column](https://dsva.slack.com/archives/G01BJ3ESXL4/p1726000369459889?thread_ts=1725993873.425839&cid=G01BJ3ESXL4) and us [updating $small-desktop-screen to 1024px](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3240) this coming sprint, here's another try: 

Screenshot 2024-09-10 at 3 51 18 PM

danbrady (approved):

micahchiang (approved):

PR #3343, opened 2024-10-07
update package versions and margin docs

micahchiang:
## Description
Updates the package versions for css-library and the margin docs with correct values for x, left, and right.


src/_data/margins.yml
micahchiang, line 1: We discovered last week that this -1px value isn't actually correct. The `vads-u-margin-x--auto, vads-u-margin-right--auto, and vads-u-margin-left--auto` classes do in fact use `auto`: 

image

jamigibbs (approved):

Issue #3054, opened 2024-07-19
Update guidance for Keep a record of submitted information pattern


caw310:
## Description
Update guidance for Keep a record of submitted information pattern. 
- Should point to confirmation page template 
- Should update to have a link to the confirmation page template. 
- This one doesn't need a new template, we just need to do a guidance change. This should point to a confirmation page template. We need a link to the confirmation page template
https://design.va.gov/patterns/help-users-to/keep-a-record-of-submitted-information

## Tasks
- [ ] Update guidance for Keep a record of submitted information
- [ ] Have DST designer and/or PO review 
- [ ] If edits are need, make those
- [ ] Publish updated guidance

## Acceptance Criteria
- [ ] guidance for Keep a record of submitted information pattern has been updated

caw310:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Design-System-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/3054?planning-poker) with Zenhub @babsdenney @danbrady @LWWright7

rachelshearerux:
👋 Hi @caw310 , [here's a direct link to the confirmation page](https://www.figma.com/design/4A3O3mVx4xDAKfHE7fPF1U/VADS-Templates%2C-Patterns%2C-and-Forms?node-id=2601-126236&t=rjIguwDRXDJ3BkbJ-1) template in the VADS Templates, Patterns, and Forms library.

LWWright7:
@jeana-adhoc  Since we are not creating a pattern "Keep a record of submitted information" would it make more sense to have the Figma link provided in the Description above point to the Confirmation page instead??

LWWright7:
@babsdenney & @danbrady - Ready for review / Pull request initiated.

PR #3354, opened 2024-10-10
Confirmation page screenshot

LWWright7:
This is updating the screenshot for the confirmation page, per ticket #3054


danbrady (changes requested):
The image is very small and gets pixelated on the site. Can we keep the new on around the same width of the one it's replacing?

PR #3361, opened 2024-10-11
Adds a better link for email templates and note about privacy officer.

humancompanion-usds:


danbrady (approved):
Approved, but confirming this is the intended rendering of the list:
image

Issue #802, opened 2022-05-04
Use Github Deployments/Environments for deploying each PR


bkjohnson:
# Description

Work for this ticket should happen after https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/793 is merged

In another ticket @humancompanion-usds [wrote](https://github.com/department-of-veterans-affairs/va.gov-team/issues/33568#issuecomment-1111083326):

> I'd really like to be able to deploy to staging and then decide when to push to production. This would give us a chance to review what we're putting out, perhaps get feedback on it.

In an effort to allow a github-centric workflow instead of running a local jekyll server to verify changes, we could use [GitHub environments for deployment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment).

For a job could have something like this:

```yaml
on: pull_request

jobs:
  dev-deploy:
    environment: 
      name: development
      url: https://dev-design.va.gov/${{ github.event.number }}
    steps:
    # ...
    - run: aws s3 sync _site s3://dev-design.va.gov/${{ github.event.number }}
```

This would allow each PR to have its own dev/preview environment that can be accessed from the PR by clicking a button. Here's an [example from `vets-website`](https://github.com/department-of-veterans-affairs/vets-website/pull/20913) showing what that would look like:

!["View deployment" button visible in PR](https://user-images.githubusercontent.com/2008881/166598651-cd6b5079-f1d4-4ef6-82fb-343aa47a3fd6.png)

## Tasks

- [ ] Create a `development` environment in GitHub
- [ ] Refactor the GitHub action for deployment to [create a reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow)
  - [ ] Remove `dev-design.va.gov` from the environment matrix
  - [ ] Update existing deploy workflow to use new reusable workflow for staging & production environments
- [ ] Create new workflow for deploying to dev
  - [ ] Should use reusable workflow
  - [ ] Deploys to dev s3 bucket with PR number as subdirectory

## Future work

An optional piece of future work could be to split up the deploys to staging & prod and:

1. Deploy to staging on a push to `master`
1. Have the prod deploy configured to [require an approval](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments) before running

jamigibbs:
[...]

PR #3435, opened 2024-10-24
Add Github preview deployment workflow

jamigibbs:
This PR will add a workflow for deploying a preview of the changes in a PR to the dev site `https://dev-design.va.gov/`. The preview will be accessible at a unique URL based on the Github PR number. 

For example, the preview link for this PR is: https://dev-design.va.gov/3435

It can also be accessed by clicking the "View deployment" button at the bottom of the PR:

Screenshot 2024-10-28 at 11 05 21 AM

related issue:

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/802

related follow-up:
- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3450


src/_includes/_side-nav.html
```diff
@@ -44,7 +44,7 @@
                               {% assign is-current = false %}
                             {% endif %}
                           <li>
-                            <a class="{% if is-current == true %}usa-current{%- endif -%}" href="{{p.permalink }}{{ item.sub-page | slugify }}">
+                            <a class="{% if is-current == true %}usa-current{%- endif -%}" href="{{site.baseurl}}{{p.permalink }}{{ item.sub-page | slugify }}">
```
jamigibbs, line 47: Sidebar sub-menu items were not receiving the preview ID so the links were pointing back to prod in the preview environment. I'm adding this `site.baseurl` so that it forms correctly _but_ we should confirm in production once the PR is merged that those sidebar links are still working as expected.

The Jekyll config file for this variable has it set to `""` in prod so I'm not expecting this to break anything there:

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/d530305882e7df3248cd1681aadb279c11ca6dfd/_config.yml#L3

In the preview workflow, we are setting this to the preview id (github event number):

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/d530305882e7df3248cd1681aadb279c11ca6dfd/.github/workflows/preview.yml#L31

README.md
```diff
@@ -90,9 +90,23 @@ While `vets-design-system-documentation` is running and make further updates to
...
+## Deployments
 
-Merges into `main` will automatically be deployed to `dev-design.va.gov`. Production is automatically deployed every weekday at 2pm. Deploys are executed by creating a release of vets-website via Jenkins. You can track the deployment in the Slack channel, #design-system.
+Merges into `main` will automatically deploy to production `design.va.gov` after CI checks have completed. 
+
+### Previewing Pull Requests
+
+When a PR is created, a preview of those changes will be available after the [Preview workflow](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/.github/workflows/preview.yml) has completed. When the workflow is complete, there will be a "View deployment" button visible that will launch the preview environment when clicked.
+
+[The view deployment button on a pull request](src/images/readme/pr-view-deployment-button.png)
```
jamigibbs, line 101: I think these images will need to be merged into `main` before I can get them to display in the README. I might have to do a follow-up PR for it.

src/index.html
```diff
@@ -6,7 +6,7 @@
 <div class="site-hero clip-bottom">
   <div class="site-l-wrapper">
       <h1 class="vads-u-measure--1 vads-u-font-family--sans">
-        A system to help you write, design, and build digital services on VA.gov
+        <span style="color: #00bde3">Jami was here</span> - A system to help you write, design, and build digital services on VA.gov
```
jamigibbs, line 9: Will remove before merging

src/_includes/_site-in-this-section.html
```diff
@@ -8,7 +8,7 @@
     <dd role="definition" class="site-in-this-section__definition">
       <a
         class="site-in-this-section__link"
-        href="{{ page.url }}{{ item.sub-page | slugify }}"
+        href="{{site.baseurl}}{{page.permalink }}{{ item.sub-page | slugify }}"
```
jamigibbs, line 11: This should be tested in prod once merged:

Screenshot 2024-10-28 at 12 02 17 PM

.github/workflows/deploy.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       run: |
         find _site/ -type f ! -iname 'index.html' -iname '*.html' \
         -print0 | while read -d $'\0' f; do mv "$f" "${f%.html}"; done
-    - uses: aws-actions/configure-aws-credentials@v1
+    - uses: aws-actions/configure-aws-credentials@v4
```
jamigibbs, line 44: I was seeing a warning about this action so I'm updating to the latest version which seems to resolve it.

Screenshot 2024-10-28 at 2 47 23 PM

.github/workflows/deploy.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   deploy:
     strategy:
       matrix:
-        environment: [{ bucket: dev-design.va.gov, config: dev.yml }, { bucket: staging-design.va.gov, config: staging.yml }, { bucket: design.va.gov, config: prod.yml }]
```
jamigibbs, line 11: Remove the dev bucket otherwise the previews will be deleted every time a PR is merged into `main`.

.github/workflows/deploy.yml
```diff
@@ -8,7 +8,7 @@ jobs:
...
+        environment: [{ bucket: staging-design.va.gov, config: staging.yml }, { bucket: design.va.gov, config: prod.yml }]
```
jamigibbs, line 11: Removing the dev bucket here otherwise the previews will be deleted every time a PR is merged into `main`

.github/workflows/preview.yml
```diff
@@ -0,0 +1,25 @@
...
+    types:
+      - opened
+      - reopened
+      - synchronize
+
+jobs:
+  dev-deploy-preview:
+    strategy:
+      matrix:
+        environment: [{ name: development, bucket: "dev-design.va.gov/${{ github.event.number }}", config: dev.yml }]
```
jamigibbs, line 14: Even though this is only deploying to a single environment, I'm using a [matrix strategy](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) to make it easier to use the reusable deployment workflow with the `deploy.yml` file which deploys to staging and production.

.github/workflows/deploy-reusable.yml
```diff
@@ -0,0 +1,84 @@
...
+        with:
+          node-version: 16.x
+          cache: 'yarn'
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.7.5 # Not needed with a .ruby-version file
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+      - run: yarn install
+      - run: yarn run build
+      - run: bundle exec jekyll build --config _config.yml,jekyll-configs/${{ inputs.matrix-config }} --baseurl /${{ inputs.github-event-number }}
```
jamigibbs, line 48: When this reusable workflow is eventually used with the `deploy.yml` workflow, I am anticipating that the `--baseurl` will receive a null value for `inputs.github-event-number` since only the preview environments need to use this. So when the jekyll build runs, the baseurl should just be set as `/` and the path won't change.

This is something that will need to be verified [when that work is done](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3450).

.github/workflows/deploy.yml
danbrady: @jamigibbs Does this mean the preview environments get deleted whenever ANY PR gets merged or just when that specific PR gets merged? Right now, after a PR gets merged into `main` the branch is automatically deleted. Deleting the preview environments seems reasonable to me as well because at that point production should match that PR. Right?

humancompanion-usds:
LGTM!!!!

.github/workflows/deploy.yml
jamigibbs: @danbrady I've observed that it's when any PR gets merged. It'll delete _all_ preview environments because the `main` branch merge will install a fresh install at the root (dev-design.va.gov) which knocks out all sub-folders (previews) below it.

We could make a separate cleanup job to remove a preview folder when a branch is merged to main. But I don't think there's any harm with keeping them there and maybe good for historical context? Kind of like how we can go back and look at old Chromatic links. But if we do feel like we need a cleanup, I can create a follow-up task.

danbrady: @jamigibbs Got it, thanks for the explanation. I didn't realize all the Chromatic links hang around too. I was concerned if there was a space limit or we'd get charged for additional space which might end up being a surprise cost, but it doesn't sound like that's the case. GTG! 🚀

micahchiang (approved):
Don't see anything that sticks out, nice work! Thanks for leaving all the comments you did.

README.md
```diff
@@ -90,9 +90,23 @@ While `vets-design-system-documentation` is running and make further updates to
...
+
+When a PR is created, a preview of those changes will be available after the [Preview workflow](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/.github/workflows/preview.yml) has completed. There will be a "View deployment" button visible that will launch the preview environment when clicked.
+
+[The "view deployment" button on a pull request](src/images/readme/pr-view-deployment-button.png)
+
+The preview link can also be accessed directly using this pattern: `https://dev-design.va.gov/PR_NUMBER`. For example, `https://dev-design.va.gov/3435` where `3435` is the unique PR number.
+
+If the preview has stopped working (for example if you're only seeing a 403 Forbidden error or similar), try re-running the preview workflow by clicking "Details" next to the Preview check and then click the "Re-run all jobs" button.
+
+[A list of PR checks with an arrowing pointing to the details link for the Preview workflow](src/images/readme/pr-checks-preview-details.png)
```
Andrew565, line 107: "arrow" instead of "arrowing"

Andrew565 (approved):
One typo otherwise good to go!

PR #3493, opened 2024-11-05
Adds missing table variation and updates the banner to point to Fall updates

humancompanion-usds:

PR #3494, opened 2024-11-06
Removing remaining alert code

danbrady:


danbrady:
This fixed is already captured in PR #3493. Closing.

PR #3504, opened 2024-11-07
Claims and applications page

megzehn:
Creating a PR for a new special topics page: claims and applications

### DIFF
--- a/src/_content-style-guide/specific-topics-and-programs/claims-and-applications.md
+++ b/src/_content-style-guide/specific-topics-and-programs/claims-and-applications.md
@@ -1,6 +1,6 @@
 ---
 layout: content-style-guide
-permalink: /content-style-guide/specific-topics-and-programs/claims
+permalink: /content-style-guide/specific-topics-and-programs/claims-and-applications
 has-parent: /content-style-guide/specific-topics-and-programs/
 title: Claims and applications
 intro-text: "The information on this page will help you use consistent language about claims and applications across communication channels."


megzehn:
Hi @humancompanion-usds, this PR is ready for the content style guide. It's the special topic page about claims and applications that we showed you on a previous design/content sync. Do you mind doing a final review and merging when you have a minute?

@DanielleThierryUSDSVA and @jstrothman both reviewed and approved it. 

Thank you!! CC @laurwill

megzehn:
Hi @danbrady, I'm pinging you since Matthew's out at the offsite. Any chance you could review and merge this PR for the content style guide?

src/_content-style-guide/specific-topics-and-programs/claims-and-applications.md
```diff
@@ -0,0 +1,59 @@
+---
+layout: content-style-guide
+permalink: /content-style-guide/specific-topics-and-programs/claims
```
danbrady, line 3: I see the page was renamed to "claims and applications", so we likely want to update the permalink as well.

danbrady (changes requested):
Since this is a new page, we'll have to add it into the left side navigation so it's accessible. Also, we would usually update the "In this section" part on the page level above (in this case, "Specific topics and programs") with a link to this new page. Are you able to do that?

megzehn:
Thank you for catching that, @danbrady!! I adjusted the permalink and added it as a sub-page on the "Specific topics and programs" index page. But I don't see where to link to it or how to add it to the nav. Could you check again and see if there are more changes I need to make?

danbrady:
@megzehn My mistake, I forgot that the left nav updates automatically based on what's in the "sub-pages" front matter. I can see it now after you added it to that section. Looks good. Approved!

danbrady (approved):

PR #3507, opened 2024-11-08
Add syntax highlighting for markdown code blocks

adamwhitlock1:
## Summary:
- Adds syntax highlighting to code blocks via [rouge gem](https://github.com/rouge-ruby/rouge)
- Adds a `copy` button to easily copy code to clipboard. This is not something we have to include, but I thought it was a nice addition. It uses the VA button and only about 35 lines of JS to get the copy functionality working.
- Syntax styling aka theme for code styling taken somewhat from some of the existing styles, but instead of using hardcoded color values was able to sass variables instead. I commented on what each class style is responsible for as far as syntax highlighting, but let me know if any of that looks incorrect or should be changed. The old styles weren't actually used at all since highlighting was not set up in the _config.yml.

## Screenshots

I went through the site to look for any styling issues that this change _could_ make and tried to address any of those issues where they occurred. There really weren't many layout problems introduced by this, but below are some screenshots of the main types of pages that have changed.

|         | Before | After |
| ------- | ------ | ----- |
| about using web components | about-using-web-components-before | about-using-web-components-after |
| flexbox grid  | flexbox-grid-before | flexbox-grid-after |
| prefill pattern | prefill-pattern-before | prefill-pattern-after |

Mobile
Screenshot 2024-11-12 at 1 26 54 PM

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -44,10 +44,6 @@ markdown: kramdown
 highlighter: rouge
 kramdown:
   syntax_highlighter: rouge
-  syntax_highlighter_opts:
-    block:
-      line_numbers: false
-      start_line: 1
 plugins:
   - jekyll-last-modified-at
   - jekyll-redirect-from
--- a/_config.yml
+++ b/_config.yml
@@ -9,7 +9,7 @@ example_library_name: "VADS Example Library"
 resource_library_name: "VADS Templates, Patterns, and Forms"
 
 # Common links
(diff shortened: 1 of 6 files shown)


adamwhitlock1:
@humancompanion-usds I'd be curious to get your opinion on this syntax highlighting as far as the color that I chose. I tried to align with the design system as much as possible and use the token colors throughout.

Andrew565 (approved):
LGTM

humancompanion-usds:
Looks great @adamwhitlock1 !

Issue #3178, opened 2024-08-12
Update guidance for process list eyebrow - documentation


caw310:
## Description
Update the examples for process list. See Slack thread for details. 
https://dsva.slack.com/archives/C01DBGX4P45/p1723483551152199

![Image](https://github.com/user-attachments/assets/767d5103-b181-4c91-ab15-07a394317fe5)

## Details
Process list guidance page [https://design.va.gov/components/process-list](https://design.va.gov/components/process-list)

## Tasks
- [ ] Update the examples and guidance for process list (eyebrow) 
- [ ] Get a DST designer to review
- [ ] Update as needed
- [ ] Publish changes

## Acceptance Criteria
- [ ] Process list has updated examples and guidance (if needed)

## Estimating
Designers: 
Provide your estimate of 1, 2, 3, 5, 8 or 13 
@babsdenney
@danbrady
@LWWright7

caw310:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Design-System-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/3178?planning-poker) with Zenhub @babsdenney @danbrady @LWWright7

LWWright7:
@danbrady & @babsdenney I created a pull request to update the screenshot for the example of education benefits (form 22-1995).  However, it looks like somebody has already updated the screenshots of the other two examples (default & status) to match the conversation that Dan & Skyler had in Slack.  

![Image](https://github.com/user-attachments/assets/5d7a0de1-9595-4737-ac3d-adba43fdbfee)

![Image](https://github.com/user-attachments/assets/cc29eefd-147d-4903-a2cb-84500d839c56)


Is this all that needs to be done??

Should anything be added to the usage section towards the bottom?? 

(Sorry, I haven't done many documentation tickets before)

LWWright7:
@babsdenney & @danbrady   What do you think of this as an addition to the "Content considerations" section of the documentation?

![Image](https://github.com/user-attachments/assets/62151b8e-e4cb-4805-ab3e-18eddcd2e56f)

LWWright7:
@danbrady & @babsdenney - New pull requests for the updated screenshot & verbiage.

PR #3555, opened 2024-12-02
Update screenshot for process list documentation

LWWright7:
Per ticket #3178 this screenshot is needed to show v3 components


danbrady (changes requested):
@LWWright7 Process list still looks like old v1 version, no? Do we need to add any documentation on what or when to use the status/eyebrow text?

LWWright7:
This is the "default" process list shown in the screenshot. The "status" version uses the eyebrow.  And yes, if you look at the [ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3178) I made a comment about whether or not I should add some verbiage to the usage section.

LWWright7:
Finding a new screenshot as this one is not accurate.

PR #3557, opened 2024-12-03
Replacing the image of the process list example

LWWright7:
For ticket 3187 (Update guidance for process list eyebrow) the example on the doc site uses the v1 component. This image is updated with v3


humancompanion-usds:
@LWWright7 - Can you view the preview of this and ensure this image shows up properly on the site now? I don't usually put spaces in the file names so I'm concerned this may not show up in the page given this new name.

LWWright7:
The previous file had spaces (that were turned to hyphens) and is like 2 miles long! lol

Screen Shot 2024-12-03 at 7 47 36 PM

And yes, when I did the preview before submitting the pull request I could see the image.  But I don't think that it will show up in the "view deployment" link above because the pull request #3558  to change the code for the page (that shows where this image is being accessed from) has not been approved yet.

Issue #3505, opened 2024-11-08
Update a11y considerations guidance for tables to discourage use of tables for non-data


humancompanion-usds:
### Duplicate check

- [x] I've searched for any related issues and avoided creating a duplicate issue.

### This update is for:

Component

### What is the name?

va-table

### What is the nature of this update?

- [ ] How to build this component/pattern
- [ ] When to use this component/pattern
- [ ] When to use something else
- [ ] Usability guidance
- [x] Accessibility considerations
- [ ] Content considerations
- [ ] Implementation
- [ ] Package information
- [ ] Addition to Forms Library documentation
- [ ] Update to existing Forms Library documentation

### What problem does this solve?

We had a Design Intent with a team who switched from a Card to a Table display for non-data and @artsymartha68 and I had to push back on them to explore other designs. Currently in Table guidance we just point to the USWDS a11y guidance. It says nothing about us not wanting teams to use tables for non-data displays/layout. We need to add our own guidance pushing teams away from this option as it is not easy for screen reader users to navigate tables for display purposes. We could offer up alternatives in the guidance (using more data-sense Card displays, offering filtering and the ability to compare Cards, etc.)

### Additional Context

I want to do this now because in the adding of new Table variations to our v3 component, some teams are taking that as a signal that we want them to use tables when we really don't unless they are displaying data.

rsmithadhoc:
PR opened here https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/3619

PR #3619, opened 2024-12-16
Update table a11y guidance

rsmithadhoc:
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3505

- Updated a11y considerations for table usage
- Kept link for USWDS' table guidance, but downgraded it to a standard link at the bottom of the list.

image

### DIFF
--- a/src/_components/table.md
+++ b/src/_components/table.md
@@ -77,4 +77,6 @@ The v3 default table is a standard table and does not offer the responsive stack
 
 ## Accessibility considerations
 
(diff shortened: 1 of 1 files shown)


danbrady (approved):

humancompanion-usds (changes requested):
Please follow the pattern we have in guidance of using the action link to USWDS and then adding our guidance underneath under the header "Additional guidance for VA". You can see an example on this same Table guidance page under Usage.

humancompanion-usds (approved):

PR #3679, opened 2025-01-09
Update Confirmation screenshots and guidance to latest template

danbrady:


humancompanion-usds (approved):
Thanks for jumping on this quickly. Good work.

PR #3761, opened 2025-01-29
Update language-toggle.md

babsdenney:
Added storybook example of language toggle.


Andrew565 (approved):

humancompanion-usds:
@babsdenney @caw310 @micahchiang - Approving this guidance update. I have a follow-on task. Can we go replace at least 1 instance (great if you want to do more) of the toggle in production with our new component and then update the screenshot in guidance so that it matches the new component? Thanks!

Issue #3716, opened 2025-01-22
Discovery: Integrating mobile components into VADS Storybook


caw310:
## Description
Currently we link to the mobile team's Storybook components from the VADS Storybook. We need to research how to integrate their Storybook components into the VADS Storybook. 

## Details
_optional, delete section if not needed_

## Tasks
- [ ] Research methods for integrating the mobile teams Storybook components into VADS
- [ ] Present options to DST engineers
- [ ] Make a recommendation for the best way 
- [ ] Draft guidelines / instructions

## Acceptance Criteria
- [ ] There is a recommended way to integrate
- [ ] Guidelines / instructions written up

jamigibbs:
Is the end goal to be able to display mobile components on design.va.gov in the same way the existing web components display for guidelines?

caw310:
Yes, the end goal is to have mobile components displayed on design.va.gov as VADS components.

jamigibbs:
Since we're already displaying [mobile components in Storybook](https://design.va.gov/storybook/?path=/docs/va-mobile_alert--docs), a possible way we could also display them in design.va.gov is to create another [storybook-preview.html](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/src/_includes/storybook-preview.html) file specific for the [mobile Storybook](https://department-of-veterans-affairs.github.io/va-mobile-library/index.html?path=/docs/design-tokens-colors--docs). 

And then just use that jeykll file when we want to add the component preview like this:

```
{% include storybook-mobile-preview.html story="va-mobile_alert--docs" link_text="Example alert link text" height="60px" %}
```

PR #3810, opened 2025-02-17
Add VA Design System Storybook link for mobile components

jamigibbs:
For both web and mobile components, all of the documentation Storybook links will point to the design.va.gov Storybook with this update (instead of the mobile Storybook instance).

Preview link: https://dev-design.va.gov/3810/components/button/

Screenshot 2025-02-17 at 1 32 52 PM

Navigates to: https://design.va.gov/storybook/?path=/story/uswds-va-button--primary

Screenshot 2025-02-17 at 1 33 12 PM

Navigates to: https://design.va.gov/storybook/?path=/story/va-mobile_button--base

related issue: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3716


src/_components/button/index.md
```diff
@@ -54,15 +54,15 @@ anchors:
 
 #### Base - Primary
 
-{% include storybook-preview.html height="100px" story="button--base" base_path=site.storybook_mobile_path link_text="va-mobile__button--base" %}
+{% include storybook-preview.html height="100px" story="button--base" link_text="va-mobile__button--base" is_mobile=true %}
```
jamigibbs, line 57: When a component is flagged as `is_mobile=true`, iframe preview and Storybook documentation links will automatically adjust depending on the component type. In both cases, the Storybook link will bring the user to the design.va.gov Storybook instance.

src/_includes/storybook-preview.html
```diff
@@ -1,16 +1,20 @@
 {% assign storybook_path = site.storybook_path %}
-{% if include.base_path %}
-  {% assign storybook_path = include.base_path %}
-{% endif %}
+{% if include.is_mobile %}{% assign storybook_path = site.storybook_mobile_path %}{% endif %}
+{% assign storybook_iframe_path = storybook_path | append: '/iframe.html?id=' | append: include.story | append: '&amp;viewMode=story' %}
+
```
jamigibbs, line 4: The main difference here between the web and mobile component iframe preview is the base link which will be from either `https://department-of-veterans-affairs.github.io/va-mobile-library/` or `https://design.va.gov`. 

**These root URLs are not clickable or displayed visually in the documentation.** They are only used to load the component in the iframe.

src/_includes/storybook-preview.html
```diff
@@ -1,16 +1,20 @@
...
     height="{{ include.height }}"
-    src="{{ storybook_path }}/iframe.html?id={{ include.story }}&amp;viewMode=story"
+    src="{{ storybook_iframe_path }}"
   ></iframe>
 </div>
 
-[{% if include.link_text %}<img aria-hidden="true" role="img" src="{{ site.baseurl }}/images/storybook.svg" class="site-component-resources-links__icon" width="16px" alt="Storybook logo">View {{ include.link_text}}{% else %}See this{% endif %} in Storybook]({{ storybook_path }}/?path=/docs/{{ include.story }})
+{% assign story_name_prefix = '' %}
+{% if include.is_mobile %}{% assign story_name_prefix = 'va-mobile_' %}{% endif %}
+{% assign storybook_preview_path = site.storybook_path | append: '/?path=/story/' | append: story_name_prefix | append: include.story %}
```
jamigibbs, line 18: The condition here is to add the `va-mobile_` prefix to the story name so that we can navigate to the design.va.gov Storybook documentation page.

src/_includes/storybook-preview.html
humancompanion-usds: Hmm. We eventually want https://department-of-veterans-affairs.github.io/va-mobile-library/ to go away and be rolled into design.va.gov. What do we need to do to not be dependent on the deployed github mobile "design system"?

jamigibbs: > What do we need to do to not be dependent on the deployed github mobile "design system"?

@humancompanion-usds I am working on a discovery for that now! :) I see this approach as a bridge until a full integration can be done. I see a few different ways we can do it.

humancompanion-usds (approved):

PR #3897, opened 2025-03-11
add storybook embed, remove unused sections

adamwhitlock1:
Description:

- Replaces the links to images for the examples to instead use the actual storybook embeds.

- Removes sections that don't have content

### DIFF
--- a/src/_components/service-list-item.md
+++ b/src/_components/service-list-item.md
@@ -28,12 +28,12 @@ anchors:
 
 ### With the optional link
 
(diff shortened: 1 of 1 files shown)


humancompanion-usds (approved):
There is an include for Code Usage that should be in the template for components. That should pull in the component's props if we've updated the component-library package in package.json.

adamwhitlock1:
> There is an include for Code Usage that should be in the template for components. That should pull in the component's props if we've updated the component-library package in package.json.

Thanks for pointing this out! I have a few more changes that are proposed to the related pages for the card component so when I make those changes I will look at using this include to get the props shown in the docs.

PR #3916, opened 2025-03-14
Update collection.md

babsdenney:
Removed the [note](https://design.va.gov/components/link/collection#instances-of-this-component-in-production) about moving this to a web component in the guidance.

Wondering if we should add an alert like the one on [Side Nav](https://design.va.gov/components/sidenav) to indicate this component isn't supported with code?

### DIFF
--- a/src/_components/link/collection.md
+++ b/src/_components/link/collection.md
@@ -14,6 +14,22 @@ anchors:
   - anchor: Accessibility considerations
 ---
 
+<va-alert
+  close-btn-aria-label="Close notification"
+  status="warning"
+  visible
+  uswds
+>
+  <h2 slot="headline">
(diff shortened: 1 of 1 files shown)


babsdenney:
@danbrady @LWWright7 Do you think we need to add a banner to this component to show this isn't a web component like we do for the [side nav](https://design.va.gov/components/sidenav)?

babsdenney:
@humancompanion-usds I talked to @micahchiang about this component. He confirmed that we aren't referencing the link collection component in the CSS Library, so I added that banner that we use on the side nav. 

@danbrady @LWWright7 This is ready for review.

LWWright7 (approved):
Looks good to me!!

Issue #3621, opened 2024-12-16
Update or swap imports for the documentation repo


caw310:
## Description
Now that we are done with formation deprecation, we need to update packages to make sure our style imports are coming from css-library. 

## Considerations
- We'll need to update [this css-library import](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/package.json#L41)
- The doc site still relies heavily on v1 grid, which we patched directly in vets-website. We'll need to migrate grid first before fully removing formation from the doc site. 
- There's extensive use of scss variables in `/assets/stylesheets/` throughout many stylesheets. It's possible there are some being used in here that we don't carry in css-library, such as `$font-bold` and `$font-sans` in the `_do-dont.scss` stylesheet. Instances like this will need to be updated to whatever their equivalents are from the variables we do generate. Running a `yarn watch` will likely throw variable errors for these once import paths are updated. 

## Tasks
- [ ] Update [these stylesheet imports to point to css-library](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/src/assets/stylesheets/application.scss#L6-L12). It's possible imports from css-library will need to include mixins, functions, and utilities.
- [ ] Update the css-library version in package.json

## Acceptance Criteria
- [ ] Import paths have been updated for the doc site so they point to css-library

powellkerry:
There is a reference to an example of how to import base css styles in `src/_foundation/spacing-units.md`, it was temporarily removed when documentation was updated to remove mention of formation. This should be re-added as part of this work as the code in the example should reflect css-library imports.

caw310:
Provide your estimate of 1, 2, 3, 5, 8 or 13 
@Andrew565 - 
@ataker - 
@harshil1793 -  
@it-harrison -
@jamigibbs -
@micahchiang -
@powellkerry -
@rmessina1010 -
@rsmithadhoc -

powellkerry:
I did a bit of work trying to see if I could quickly switch this out for testing out changes to list styles. See this [branch](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/tree/3621-replace-formation).

powellkerry:
This PR should be merged after this work is complete: https://github.com/department-of-veterans-affairs/component-library/pull/1508

PR #3933, opened 2025-03-20
Replace Formation with CSS-Library

jamigibbs:
This PR will:

- Remove the deprecated Formation library
- Import necessary CSS-Library files
- Update Font Awesome icons to va-icon web component
- Replace action link classes with the va-link-action web component
- update tables with va-table web component
- [migrate deprecated grid classes](https://vfs.atlassian.net/wiki/spaces/DST/pages/3968991267/v1+to+v3+grid+migration+guidance)

closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3621


config/gulp/css.js
```diff
@@ -1,12 +1,19 @@
...
-  return stream;
+gulp.task('clean-css-library-asset-paths', function() {
+  console.log('cleaning css-library asset paths');
+  // Removing the ~ prefix from the asset paths
+  return gulp.src([
+    './node_modules/@department-of-veterans-affairs/css-library/dist/stylesheets/**/*.css',
+    './node_modules/@department-of-veterans-affairs/css-library/dist/stylesheets/**/*.scss'
+  ])
+    .pipe(replace(/~@uswds/g, '@uswds'))
+    .pipe(replace(/~@department-of-veterans-affairs/g, '@department-of-veterans-affairs'))
```
jamigibbs, line 13: This is where things get weird. The `~` we use in the CSS-Library to reference assets in other NPM packages (fonts and images) does not work with Jekyll because Jekyll doesn't use webpack or another Javascript bundler that tells it to "resolve from node_modules". 

I tried to spoof this link by moving the fonts and images into a folder like `/~uswds/uswds/etc` but in the browser, the tilde `~` [doesn't load properly in the browser](https://jkorpela.fi/tilde.html) either.

So this approach is to remove that character from the CSS-Library stylesheets and add the fonts and images to folders named `/@department-of-veterans-affairs` and `/@uswds` (without the `~`).

_config.yml
```diff
@@ -51,7 +51,6 @@ source: src
 sass:
   sass_dir: assets/stylesheets/
   style: compressed
-  implementation: sass-embedded
```
jamigibbs, line 54: This doesn't seem to be a configuration option that exists anymore so removing it: https://github.com/jekyll/jekyll-sass-converter?tab=readme-ov-file#configuration-options

src/_about/maturity-scale.md
jamigibbs, line 1: **before**

Screenshot 2025-03-20 at 4 48 04 PM


**after**

https://dev-design.va.gov/3933/about/maturity-scale

Screenshot 2025-03-20 at 4 47 05 PM

src/assets/javascripts/sidenav-formation.min.js
jamigibbs, line 1: This file was originally named `formation.js` but it is the script that handles the sidenav for the site so I'm keeping it here but just with a new name so we know why it's still here.

If we'd like to completely deprecate this script, we might want to create a separate ticket to re-build the side nav:

Screenshot 2025-03-21 at 10 46 22 AM

src/_about/maturity-scale.md
jamigibbs: @danbrady @babsdenney @LWWright7 We replaced Font Awesome icons with va-icon and made some best guesses on how to map a few that weren't a 1:1 swap. Could you let us know if you have other preferences for these?

config/gulp/css.js
humancompanion-usds: Is this a Jekyll problem or a CSS problem? Would this be fixed by us updating to dart-sass in this repo? I guess I'm wondering if we don't go forth with the Node upgrade in here in the not too distant future and get off of node-sass entirely. I would presume then we could use webpack or another bundler if we wanted to?

humancompanion-usds (approved):
This really great and thorough. I like the icon choices. The only one I had a question about is the + and - for the accordions in the sidenav. It still looks chonky compared to the icons used in va-accordion. That's my only quibble. Thanks for taking care of this!!

This also pointed out to me that our sidenav both here and on VA.gov is really...dubious. It overrides the font-size to 15px for the most part but not always which is causing some shifting of text when you interact with it. But, problem for another day.

config/gulp/css.js
jamigibbs: @humancompanion-usds This seems to be a limitation with our current Jekyll setup. We could try to integrate Webpack as part of the build (I like that idea actually) but Jekyll doesn't natively do it.

What I'm reading about the [Jekyll Sass Converter](https://github.com/jekyll/jekyll-sass-converter?tab=readme-ov-file#usage) is that it gets bundled with Jekyll and uses `sass-embedded` which in turn leverages the `dart-sass` compiler so it doesn't appear to be an issue related to that.

jamigibbs:
> The only one I had a question about is the + and - for the accordions in the sidenav. It still looks chonky compared to the icons used in va-accordion.

@humancompanion-usds Good catch. We should update those with va-icon too.

> This also pointed out to me that our sidenav both here and on VA.gov is really...dubious. It overrides the font-size to 15px for the most part but not always which is causing some shifting of text when you interact with it. But, problem for another day.

Yes, technically the "sidenav" uses the `.usa-accordion` classes that originated from Formation. It probably needs to be rebuilt.

Andrew565 (approved):

danbrady:
The action links are showing inline on [this page](https://dev-design.va.gov/3933/foundation/design-tokens):

image

I'm not sure if this is an issue with the removal of Formation or with `va-action-link` itself?

danbrady:
It looks like our flexbox layouts were affected:

Left = Live
Right = Preview env of PR 

image

danbrady:
I used a visual diff tool ([backstopJS](https://github.com/garris/BackstopJS)) and every page I tested always showed the main content shifted down a bit:
image

In the preview environment, the  `<main>` tag has a `large-screen:vads-u-margin-top--neg2` class but it doesn't have any affect. However, if I remove the responsive prefix to `vads-u-margin-top--neg2` does.

So perhaps our responsive prefixes may not be working?

powellkerry:
@danbrady that was a good find, we had updated all the breakpoint class names for grid but didn't do it for the other utility classes. Should be fixed now.

PR #3970, opened 2025-04-01
What's New: Updates for v0.0.438

danbrady:
[Open Preview Environment](https://dev-design.va.gov/3970)


caw310 (approved):

jamigibbs:
@danbrady Is it possible to manually add @powellkerry too to this update?

Screenshot 2025-04-01 at 3 26 30 PM

danbrady:
Definitely. I manually edited the release feed we get from GH to add Kerry. There might be an "official" way to have multiple authors credited in a PR [according to this](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors), which I didn't realize was possible. It looks like if we include the co-authors name within a commit message, they'll be credited in the release notes. Worth a shot next time!

PR #4011, opened 2025-04-15
Breakpoints: Add Sass variables to the breakpoint table

jamigibbs:
## Summary

Media queries don't seem to support custom properties. For example:

```
@media (max-width: var(--medium-screen)) 
```

So it would probably be helpful if the Sass version of the breakpoints were added to the breakpoint table.

```
@media (max-width: $medium-screen) 
```

**Before**

Screenshot 2025-04-15 at 9 28 37 AM

**After**

Screenshot 2025-04-18 at 9 48 16 AM

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4011)


src/_foundation/breakpoints.md
```diff
@@ -71,7 +67,7 @@ You may place the `@media` mixin inside of a selector or places selectors inside
     margin: units(2);
   }
 
-  @include media($large-screen) {
+  @include media($desktop-lg) {
```
humancompanion-usds, line 70: We have $large-screen in some other scss files as well. Do you mind tackling those or filing an issue for us to do that? I'm curious, did we leave the Formation variable names behind and thus we "grandfathered" in or are these older media query variables broken at the moment? Just wondering how much urgency we need to have to address them.

humancompanion-usds (approved):

src/_foundation/breakpoints.md
jamigibbs: The old Formation breakpoint names are still generating in CSS-Library so technically still available while teams transition. I'll knock out the ones here though. A migration script for vets-website might be helpful. 🤔 

https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/css-library/tokens/breakpoints.json

jamigibbs: I gave this a little more thought and I think we should keep the column with the old VADS names while we are still needing to migrate old VADS names. I added it back but with "Deprecated" in the header. There are many instances still in vets-website and it would be confusing trying to map to the new name without it in the table still.

PR #4052, opened 2025-04-28
Icons: Update guidance and alphabetize icons

### LINKED ISSUE #2930: Guidance / Update Icon page layout
### Duplicate check

### This update is for:

Component

### What is the name?

va-icon

- [X] When to use this component/pattern
- [ ] Usability guidance
- [ ] Accessibility considerations
- [ ] Content considerations
- [ ] Implementation
- [ ] Package information

The current foundation icon page uses categories in two ways: where the icon is used (ex. Messaging) and what type of icon it is (ex. Social). Also, sometimes an icon can logically fit into two different categories. This may lead to confusion on how to categorize icons and how to find the right one for a usage.

We need to restructure the layout of icon page to be easily scannable and searchable. Some research will be needed to determine the best layout. Open questions:

* Should we alphabetize the icons?
* Should we add other keywords for searchability? (Like we do in Figma.)
* Can we remove the reference to FA icons?
* Should we separate "Usage" into how and where it's used?
* Should foundation and component icon pages be combined to ease finding the right guidance?

### Additional Context

_No response_

danbrady:
## Related Issue

Closes #2930 

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4052)


humancompanion-usds (commented):
I don't love that we're losing the IA in that we don't have a view now that shows all of the icons organized by section. Are we sure this is what VFS teams want? Or is this just easier for us to maintain?

danbrady:
@humancompanion-usds When the original issue was created, we discussed how the current icon groupings introduce several quirks:

* The categories mix *where* icons are used (Appointments, Profile, developer.va.gov, etc) with *how* they're used (Navigation, Interaction, etc.), which creates inconsistency.

* Some icons appear in multiple sections. For example, the “phone” icon is under both Appointments and Identification. I would suspect other icons show up in multiple places and are not captured in different sections on the doc site.

* If a user didn't know where the icon is used, how would they know what section to look at? For example, if the phone icon is needed, I would think Interaction or Navigation are the logical sections to look, not Appointments or Identification.

* Some icon categorization choices are unclear. For example, the YouTube logo appears under Navigation (possibly because of the va-icon variant?) and not under Social alongside other platforms icons.

* These inconsistencies can make it harder for VFS teams to locate the icons they need, especially if they don’t already know which section to look in.

Given those issues, we opted for alphabetical ordering to simplify the experience (as well as maintenance) and bring it in line with how other design systems (specifically [USWDS Icon](https://designsystem.digital.gov/components/icon/)) present their icons. (It would also be great to have a search in our storybook, similar to USWDS, to help find icons.) 

That said, I’m open to alternatives. We could run a quick poll in Slack or check with a few VFS teams to see how they search for icons?

humancompanion-usds (approved):

danbrady:
I really liked the idea of providing both alphabetical and category views so I built it! I'm using radio buttons for now until we have the segmented button or tabs ready.

[Open Preview Environment](https://dev-design.va.gov/4052/components/icon#preview) 

I also took a pass through the components library and identified all the icons that we use and grouped them in a "DS Components" category. 

Note that an icon can be part of many categories. For example "youtube" is in Navigation, DS Components, Social categories. I left the other categories as we previously had them. Like you said, we should revisit the IA at some point. Let me know what you think.

humancompanion-usds (approved):
Awesome! Well done. This is great.

PR #4062, opened 2025-04-29
Add command to map Storybook to port 8080 for local development

jamigibbs:
## Summary

Previewing components through the Storybook iframe can be done two different ways and they each have a separate port number:

1. Running Storybook using the [local developments steps](https://github.com/department-of-veterans-affairs/component-library?tab=readme-ov-file#running-build-via-storybook) (port 6006)
2. Running Storybook on a static server like outlined in [step 7 of the README](https://github.com/department-of-veterans-affairs/vets-design-system-documentation?tab=readme-ov-file#to-run-locally) (port 8080)

It seems like for some people the first option works better than the second option or the other way around so this PR is adding a script that allows someone to use either approach.

`yarn start` = port 6006
`yarn start:storybook-static` = port 8080

Screenshot 2025-04-29 at 4 34 35 PM

## Related Issue

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4062)


README.md
```diff
@@ -83,6 +91,12 @@ Tested using `node` v18.19.0 and `yarn` v1.22.19.
 
 5. **Build and start the Jekyll server**
 
+   ```
+   $ yarn start:storybook-static
+   ```
```
jamigibbs, line 96: My IDE formatted on save so there's a bunch of unrelated formatting changes. :/ 

The main bit I'm adding is the new script command which will connect Storybook from port 8080.

README.md
```diff
@@ -83,6 +91,12 @@ Tested using `node` v18.19.0 and `yarn` v1.22.19.
...
+
+   or, if not using step 7 below:
+
    ```
    $ yarn start
```
jamigibbs, line 101: This command will still exist for anyone that still wants to load the component previews from their Storybook running on port 6060.

jamigibbs:
> As for your editor, it looks like it's going with spaces instead of tabs, which is proper and correct so 👍 Also, looks like it might be enforcing markdown formatting? Can we get both setup with the appropriate dotfiles? I would appreciate that. I'm not sure I'm using the same VS Code settings as everyone else and I feel like these should be solved problems. Not necessary for this PR but would be good to do at some point.

It looks like I had my editor set to format on save the entire file instead of just the what I modified:

Screenshot 2025-05-02 at 10 33 34 AM

I'm going to revert my change for now and create a follow-up for us to look at adding a [Prettier config](https://prettier.io/docs/configuration) file similar to [what that the component-library has](https://github.com/department-of-veterans-affairs/component-library/blob/main/.prettierrc.json). We will probably want to come to some group consensus about formatting and it might be a little disruptive tracking changes once we add it but we should definitely get one added.

Once a Prettier config is available, everyone should set that as their default formatter:

Screenshot 2025-05-02 at 10 40 13 AM

jamigibbs:
Follow up ticket:

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/4069

PR #4099, opened 2025-05-09
Adds guidance for search filter.

humancompanion-usds:
## Summary

_Provide a brief summary of what this PR does, if the title doesn't fully explain it._

## Related Issue

Closes #[Issue_number]

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4099)

### DIFF
--- a/src/_components/search-filter.md
+++ b/src/_components/search-filter.md
@@ -1,0 +1,54 @@
+---
+layout: component
+title: Search Filter
+intro-text: This component implements faceted search.
+research-title: Use this to match the label in the research repo. Only use if web-component does not match the label.
+status: use-with-caution-available
+web-component: va-search-filter
+anchors:
+  - anchor: Examples
+  - anchor: Usage
+  - anchor: Behavior
+  - anchor: Code usage
+  - anchor: Component checklist
+---
+
+## Examples
+
+### Default
+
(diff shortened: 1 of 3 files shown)


src/_components/search-filter.md
```diff
@@ -0,0 +1,54 @@
+---
+layout: component
+title: Search Filter
+intro-text: This component implements faceted search.
+research-title: Use this to match the label in the research repo. Only use if web-component does not match the label.
```
danbrady, line 5: Need to update or remove this link.

src/_components/search-filter.md
```diff
@@ -0,0 +1,54 @@
...
+status: use-with-caution-available
+web-component: va-search-filter
+anchors:
+  - anchor: Examples
+  - anchor: Usage
+  - anchor: Behavior
+  - anchor: Code usage
```
danbrady, line 12: Doesn't look like there is a code usage section?

danbrady (approved):
Just a few links to update needed.

src/_components/search-filter.md
humancompanion-usds: Yeah, why not!?! There should be automatically. That whole process is opaque to me. I'll ask in Slack and add it later.

humancompanion-usds: Actually, I'm going to leave it in because it should appear once the new component is in the appropriate data file. The include is there and that contains the header for that section. We need to automate the move of that file when component-library is updated.

PR #4121, opened 2025-05-14
Maturity status: Automates sync from component-library [Fixes #3906]

### LINKED ISSUE #3906: Sync Maturity Levels - Integrate with Docs site
## Description
Using the *.tokens.json file created in [this ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3905), import the data into the Docs site and modify/create templates in order to pull maturity level data from the tokens file.

## Details
Format information for the *.tokens.json file can be found [here](https://tr.designtokens.org/format/#design-token-0). Data should be placed in [this folder](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/tree/main/src/_data) in order to get 'picked up' by the Jekyll templates.

## Tasks
- [ ] Modify and/or create templates in the Docs site that can use the *.tokens.json data

## Acceptance Criteria
- [ ] The Docs site displays maturity level data for each component, sourced from the *.tokens.json file

humancompanion-usds:
## Summary

Adds a build task to create a smaller json file from component-docs.json, which comes from component-library. Updates sidenav and status badge to use that new file. Removes statuses for web-components while retaining using status for non-web-components.

## Related Issue

Closes #[3906](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3906)

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4121)


danbrady (approved):
This is working great for the doc site, so I'm good with approving it to fix that up. There's two things that I can think of that will need further research in order to have it update Figma:
* The JSON doesn't capture the maturity of non-web-components. They still need to have their front matter provided separately. (Perhaps this front matter can be pulled when then JSON is made?) With the current JSON though, we'd need a different way to update these components in Figma.
* We're very limited in Figma with what we can do with text that the JSON brings in. We can't perform any logic to transform `best_practice` to be `Best Practice`, to my knowledge. There may be some ways around this, but likely we'll have to massage the JSON a bit to figure out the right values for both channels.

Note: While the new gulp task code looks ok to me, I think one of the @department-of-veterans-affairs/platform-design-system-fe should validate.

config/gulp/json.js
```diff
@@ -13,7 +14,38 @@ gulp.task('move-json-data', function (done) {
...
+  const docs = JSON.parse(fs.readFileSync(inputPath, 'utf8'));
+  const result = {};
+  (docs.components || []).forEach(component => {
+    if (!component.tag || !component.docsTags) return;
+    let maturityCategory = null;
+    let maturityLevel = null;
+    component.docsTags.forEach(tag => {
+      if (tag.name === 'maturityCategory') maturityCategory = tag.text;
+      if (tag.name === 'maturityLevel') maturityLevel = tag.text;
+    });
```
jamigibbs, line 36: I think this logic is fine for what it's doing and I'm ok if it's merged as-is but I would probably prefer to see a more modern functional approach using `.filter` and `.reduce` instead of nested `forEach` statements. Something like:


```js
    (docs.components || [])
      .filter(component => component.tag && component.docsTags)
      .forEach(component => {
        // Extract maturity data in a single pass
        const maturityData = component.docsTags.reduce((data, tag) => {
          if (tag.name === 'maturityCategory') data.maturityCategory = tag.text;
          if (tag.name === 'maturityLevel') data.maturityLevel = tag.text;
          return data;
        }, {});
```

jamigibbs (approved):

PR #4137, opened 2025-05-15
Destigmatizing language for weight: Fixing link

laurenvocke-va:
## Summary

Fixing the first link under Additional resources. It wasn't linking in the live page although the preview looked okay; this may have been because of the vertical bar in the title, which is now a comma.


laurenvocke-va:
@DanielleThierryUSDSVA This PR is ready for your review. I fixed the first link under Additional resources - changed the vertical bar to a comma. The original link looked okay in the preview, but on the live page it isn't linked and the formatting is off. The issue most likely stems from that vertical bar so replacing it with the comma, I hope, should fix the issue.

danbrady (approved):

PR #4152, opened 2025-05-16
File input: Added storybook examples for file input

### LINKED ISSUE #3002: Storybook - Configure File Input examples
## Description
The [Storybook examples for the new file input](https://design.va.gov/components/form/file-input) require you to upload a document to see some of the variations in the examples. We would like to make the Storybook show the end state of the example.

## Tasks
- [ ] Change the states of the Storybook examples for file input to show the variation being described

## Acceptance Criteria
- [ ] Review with PO and Designers

babsdenney:
## Summary

- Replaced images  for single file uploaded and multiple file uploaded with storybook examples
  - Multiple files uploaded example
  - Single file upload example
- Removes some duplicate examples for multiple files

## Related Issue

Closes #3002 

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4152)

### DIFF
--- a/src/_components/form/file-input.md
+++ b/src/_components/form/file-input.md
@@ -32,15 +32,11 @@ NOTE: Storybook does not allow for showing components in a particular state. Thu
 
 ### Default - Single file - Populated
 
-{% include component-example.html alt="Populated file upload with a single file" file="/images/components/file-input/file-input-default-populated-single.png" caption="File input component (uswds-va-file-input--default) showing a file attached for upload." class="x2" reverse="true" %}
-
-### Default - Multiple files - Populated singular
-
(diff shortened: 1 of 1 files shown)


danbrady:
Thanks for taking care of this, Barb. I see the "Default - Multiple files - Populated singular" example was removed, but it's not actually a duplicate. Was that intended? I think its intent was to show what a Multiple File input looks like, with only one file attached, but still has a "drag" area for additional files. That said, it is very similar to the "Default - Multiple files - Populated multiple", so I'm not oppose to removing the first one. Thanks!

danbrady (approved):

babsdenney:
@danbrady That was my thought too, that it is very similar to the example with multiple files. Now that we have an example of the multiple files, I think it would be confusing if we were showing the single file one with a screenshot and the one with multiple files as a interactive prototype. Now, you can use the interactive one to remove one of the files and it would look like the single file example as well. Do you want me to put back the single file example screenshot?

PR #4190, opened 2025-05-22
Action Links: Minor formatting for readability

danbrady:
## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4190)

### DIFF
--- a/src/_components/link/action.md
+++ b/src/_components/link/action.md
@@ -38,17 +38,17 @@ anchors:
 
 The action link is an eye-catching link to start a digital service. An action link entices users to take action. Example: Starting a benefit application.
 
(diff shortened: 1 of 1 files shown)


humancompanion-usds (commented):
Why are we renaming the variations in a non-standard way? I agree with putting the links-vs-buttons include back (not sure why that got removed; if I did that it was a mistake), but I deliberately called the component variations what they are titled. We should define component naming in a shared Copilot prompt.

danbrady:
Yeah, we should probably define a standard way to refer to components. I think part of the issue is that on the doc site and storybook, we organize components by category and then variation, like "Button - Segmented", but when we talk about them, we tend to use more natural language, like "Segmented Button."

We’re already doing that kind of thing on the current Button page:

> Avoid using many primary buttons on a single page or section.

> Use Primary for the most important action.

> Use big primary buttons for the only action.

While reviewing the Action Link guidance, I found it a bit hard to read when every variation was written out in full, like "Link - Action - Primary entry," "Link - Action - Primary," "Link - Action - Secondary." My intent was to simplify that by just referring to the variation name, since we’re already on the Action Link page and the base component felt implied. But I realize now that might just be my personal take.

I also didn’t realize at the time that this approach doesn’t match how the Storybook examples are named. So feel free to close this PR if it’s not working. No offense taken or meant! 😄  

That said, maybe it’s worth running the component pages through Copilot to help bring more consistency to naming and tone across the board?

PR #4295, opened 2025-06-10
Update button-labels.md: Removes inaccurate examples 

laurenvocke-va:
## Summary

Removed the examples at the end that fit links more than buttons. Combined the two like this/not this examples so they wouldn't split the bulleted list

## Related Issue

Issue 2013


laurenvocke-va:
@DanielleThierryUSDSVA This PR for the buttons guidance is ready for review. Both the style guide and components pages pull from this button page, so this should update in both areas. I removed the examples at the end that were more accurate for links and I also combined the two like this/not this examples so they wouldn't split the bulleted list. Thanks!

humancompanion-usds (approved):

PR #4336, opened 2025-06-16
Makes the banner always point to the latest release(s).

humancompanion-usds:
## Summary

Changes the site banner to always pull in the latest release(s). 

## Related Issue

I had this idea today that we could now pull this data up to the home page and hopefully do a better job of highlighting that this is all automagical now.

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4336)


src/_includes/_banner-whats-new.html
```diff
@@ -1,9 +1,13 @@
-<div class="vads-u-background-color--success-darker vads-u-padding-y--1">
+{% assign component-library = site.data.component_library_releases[0] %}
```
jamigibbs, line 1: **question (non-blocking)**: Could you just do this instead of using the index?

```
{% assign component-library = site.data.component_library_releases.first %}
```

**thought**: By using index `0` (or the first array item blindly), we're assuming that the latest release is always going to be the first item in the array but are we guaranteeing that? It's populated by a Github API response that's probably making it possible for the first "release" not to be the official latest release (like if we create a draft or release candidate of the `component-library`).

If we wanted to guarantee the first array item was the latest, we will probably want to update the [fetching script](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/main/src/_plugins/jekyll_get_json.rb) to get the release that's tagged as **latest** and then put that at the top. Just food for thought.

Ideally we could do something like this then:

```
{% assign latest_release = site.data.component_library_releases | where: "is_latest", true | first %}
```

Claude 3.7 is giving me something like this as a possible solution using this prompt:

> Is it possible to update the script at src/_plugins/jekyll_get_json.rb so that it includes a latest flag from the github api?
 
(not tested so I'm not sure if it even works and maybe it can be done in a single API call):

Screenshot 2025-06-17 at 10 32 54 AM

jamigibbs (approved):

src/_includes/_banner-whats-new.html
danbrady: FWIW, whenever we pull the latest release notes, I check that they populate the What's new page correctly. Currently, we pull the three most recent releases. Based on the [github api docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28), it doesn't look like preleases or drafts can be set to latest:

> Drafts and prereleases cannot be set as latest. 

However, if we did a release candidate or beta, like v50.0.0-beta, that might come through the feed. Not sure how we can prevent that without doing a little more research. The [API can fetch](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release--parameters) what's tagged as the latest release, so that might be a good place to start if we decide to look into this further.

PR #4633, opened 2025-08-07
Update numbers.md: Adding ampersands and revised 1 vs one

laurenvocke-va:
## Summary

Added the ampersand information (moved from the Punctuation section). Revised the guidance on using "1" versus "one" to only use "one" with certain phrases. Made edits to align the rest of the content with our style.

## Preview Environment Links


DanielleThierryUSDSVA (approved):
@laurenvocke-va I made a few minor tweaks and I think this is good to go. Thank you!

PR #4766, opened 2025-09-03
Button: Updating urls in guidance

babsdenney:
## Summary

Updated links on when you should choose a link vs. a button

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/4766)


danbrady (approved):
@babsdenney Links look good and are working now. 🎉 

I notice that the text (that already existed) is not consistent with capitalization of "link" and "button". Since they're both components now, I believe we should be capitalizing both in both links? Also, the actual link text states "links vs. buttons" but the section header is "button vs. link". I think either way is fine, but we might want to be consistent.

PR #4842, opened 2025-09-17
Updating email guidance

derekwang99:
## Summary

_Provide a brief summary of what this PR does, if the title doesn't fully explain it._

## Related Issue

Closes #[Issue_number]

## Preview Environment Links


derekwang99:
Note: One area of concern is **Action Link > "Don’t include more than 3 Action Links total–it is strongly encouraged to only use one**". VADS guidance (outside of emails) suggests that only 1 Action Link be used. However, in the context of emails, Action Links are taking over the primary CTA function of links, which suggests that there can be at most 3, so I wrote in that 3 Action Links could be used (since there has been cases where 2 were needed).

danbrady (approved):

PR #4879, opened 2025-09-25
Search: Trying out pagefind for search

humancompanion-usds:
## Summary

An attempt at improving our search on design.va.gov. Just saving this in case I want to go back to it. The results were underwhelming.

## Related Issue

No issue has been filed but we have negative feedback about our search from users.

## Preview Environment Links


humancompanion-usds:
Closing this. It was an experiment but we're more likely to move to docusaurus at this point and using a different search solution entirely.

PR #5037, opened 2025-10-20
Enable manual deployment workflow trigger

humancompanion-usds:
Add workflow_dispatch to the Deploy site workflow to allow manual triggering from the GitHub UI.

## Changes
- Added `workflow_dispatch` trigger to the Deploy site workflow
- Added optional 'reason' input parameter for better tracking and documentation

## How to use
Once merged, you can trigger deployments manually by:
1. Going to Actions > Deploy site workflow
2. Clicking 'Run workflow' button
3. Optionally entering a reason for the deployment
4. Selecting the main branch and clicking 'Run workflow'

This provides a way to deploy the current site without needing to push new commits to main.
[Open Preview Environment](https://dev-design.va.gov/5037)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,11 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      reason:
+        description: 'Reason for manual deployment'
+        required: false
+        default: 'Manual deployment triggered'
 
 jobs:
   deploy:


.github/workflows/deploy.yml
```diff
@@ -3,6 +3,12 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: 'Reason for manual deployment'
```
jamigibbs: > description: 'Reason for manual deployment'

**recommend**: I don't think a description is required so maybe just remove.

jamigibbs (approved):

.github/workflows/deploy.yml
humancompanion-usds: Oh, I added one because it otherwise looked like a bug to me with just "workflow_dispatch:" and nothing following it. I know that works, but it looks like nothing is defined. Well, not only looks like but is in fact true.

PR #5103, opened 2025-10-30
Update accessibility issue template severity options

jeana-adhoc:
## Summary

Update accessibility issue template severity options

## Related Issue

N/A

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/5103)


amyleadem (approved):
LGTM! One small note is that we linkto the rubric twice in the template, but I don't personally mind it being very present

PR #5136, opened 2025-11-06
chore: update releases for Sprint 14

humancompanion-usds:
## Automated Release Update

This PR contains automated updates to:
- Component library releases
- design.va.gov releases

**Auto-generated by the Update Releases workflow**
[Open Preview Environment](https://dev-design.va.gov/5136)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ src/assets/stylesheets/mobile-typography.css
 src/assets/stylesheets/shame.css
 src/assets/stylesheets/uswds-typography.css
 src/assets/stylesheets/utilities.css
+src/assets/stylesheets/@*
 src/formation.min.css
 src/vendor
 src/_data/component-docs.json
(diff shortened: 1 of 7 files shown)


danbrady:
@humancompanion-usds Weird... this is the first time I'm ever seeing those in the change files list. I tested this many o' times too. But yeah, I don't think we don't need those. I'll look into why those are getting generated.

danbrady:
@humancompanion-usds I've figured out what's happening:

* During the build process, a Gulp task copies images from another repository dependency.
* The original images in that repo have incorrect permissions set.
* When our workflow runs and copies these files, it updates their permissions to remove executable bits (likely as a standard security measure).
* That's why we're seeing permission changes in this PR rather than new files being added.

Going forward, we've updated `.gitignore` to exclude all assets copied from other repositories. Since image files should not be executable, I'll keep those changes in this PR. Future PRs that run this workflow shouldn't include these files.

danbrady (approved):

PR #5143, opened 2025-11-06
Ignore assets copied from other repos

danbrady:
## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/5143)


jamigibbs:
> I ended up adding an rule to ignore any files from repos (`@*`). But I'm not sure if that may be too greedy?

I think that'll be fine. We don't necessarily need the uswds package assets checked-in either:

Screenshot 2025-11-06 at 11 32 25 AM

PR #5668, opened 2026-02-06
Footer - Minimal - Addresses staging review findings for content

### LINKED ISSUE #5667: Staging Review - Content - Design System Staging review: Minimal Footer
## VFS actions
- [ ] Assign this ticket to the team member(s) responsible for addressing feedback provided by Platform.
- [ ] Comment on this ticket:
  - [ ] If the Platform reviewer has any **Thoughts/Questions** that require responses.
  - [ ] When **Must** feedback has been incorporated. As appropriate, link to any other GitHub issues or PRs related to this feedback.
  - [ ] When **Should/Consider** feedback has been incorporated, or if any feedback will not be addressed. As appropriate, link to any other GitHub issues or PRs related to this feedback.
- [ ] Close the ticket when all feedback has been addressed.

## Thoughts
Overall, this looks pretty good. Beyond a typo, my content suggestions are to enhance the content so it's extra clear to the Design System user.

### Must
- In Storybook, a link reads, “View guidance for the Minimal footer component in the Design System.” This link is broken. Please revise by fixing the link.
- Usage > When to use footer - minimal > To reduce the chance... - Under this bullet, the user finds the following sentence: "However, it removes elements of the standard footer, such additional navigation, in order to reduce the chances of users leaving..." There is an "as" missing after "such." Please revise so the sentence reads as: "However, it removes elements of the standard footer, such as additional navigation, in order to reduce the chances of users leaving..."
- Usage > When to use footer - minimal - The user finds a bullet that reads: “A tasks that is time-sensitive.” There is a typo: the sentence should read either as “A task that is time-sensitive” or “Tasks that are time-sensitive.” Please revise.
- Footer parent page > Usage > How this component works - The user finds a section that reads, "The footer is made up of four parts," but then finds a list with only three items. Please revise by either changing "four parts" to "three parts," or add an additional item to the list.

### Should:
- The user finds the following subheading at the top of the page: “The minimal footer can be used when navigating away would prevent the user from easily accomplishing their main task.” This subheading should be rewritten so it has more context and isn’t in passive voice. There are several ways that the writer could do this. For example, they could write: “You can use the minimal header when navigating away from a form or process would prevent the user from easily accomplishing their main task.”
- Usage > When to use Footer - Minimal - All bolded text options are awkwardly worded; they’re fragments and inconsistent in formatting. They should be rewritten so they’re complete sentences and demonstrate when a user would use Footer - Minimal. There are several ways the writer could do this. For example, the bolded text options could be:
  - To reduce the chance…
  - When a task is time-sensitive.
  - When an experience is outside the VA.gov…

## Notes
Thank you for the opportunity to review this work. If you have any questions, please let me know.

humancompanion-usds:
## Summary

Addresses should recommendations from staging review of the Footer - Minimal.

## Related Issue

Closes #5667 

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/5668)

### DIFF
--- a/.github/ISSUE_TEMPLATE/6_bug.yml
+++ b/.github/ISSUE_TEMPLATE/6_bug.yml
@@ -1,4 +1,4 @@
-name: "1. 🔴 Bug report"
+name: "6. 🔴 Bug report"
 description: "Use this template to report any defects for VA Design System Components and Patterns"
 title: "[Bug]: Description of Bug"
 labels: ["bug", "platform-design-system-team"]
--- a/src/_components/alert/index.md
+++ b/src/_components/alert/index.md
@@ -219,7 +219,7 @@ Use **warning** when something *might* go wrong or requires attention before pro
 
 #### Handling multiple alerts
 
-Displaying multiple alerts should be avoided. However, when you need to display multiple alerts on a page, follow the general alert usage guidance above (including ordering alerts by severity and avoiding stacking alerts of the same type). In addition:
+Displaying multiple alerts should be avoided. However, when you need to display multiple alerts on a page, follow the general alert usage guidance above (including ordering alerts by severity and avoiding stacking alerts). In addition:
 
 * **Create visual hierarchy.** When multiple alerts are necessary, use a Standard alert for the primary message and Slim alerts for secondary messages.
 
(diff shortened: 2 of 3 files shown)


jeana-adhoc (approved):
LGTM

PR #5803, opened 2026-02-26
Update payments-and-debts.md: Updating balance definition

laurenvocke-va:
## Summary

Updating the definition for "balance"

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/5803)


laurenvocke-va:
Hi @humancompanion-usds! Not sure why, but it looks like this PR didn't update the page in the style guide. I asked some other members of my team to double check in case it was an issue on my end, but we're not seeing the updates. Do I need to create a new PR for this?

Issue #1785, opened 2023-05-24
Experimental Design [Plain divider]


aagosto90:
## What
> I'm proposing a plain hairline divider to be added to the design system as a variation of the only existing current [divider component](https://design.va.gov/components/divider).

## Purpose
> The divider can be useful in situations where content is related but separate, which is a problem that is not currently solved by the divider with stars. That component is to be used for separate sections that don't directly relate to one another, so I see a need for a plain divider.

## Usage
> **Context or task**: This divider would be used on pages where including a divider can help to organize and divide sections of content, but the sections still need to appear to be related and not completely different topics.

## Behavior
> Needs to responsively resize horizontally, but there is no interaction with this component.

## Examples
> I realized this component doesn't yet exist when designing a notification center MVP - a list page of past notifications (not live yet). Here's an example of how this component could be utilized in a list:
<img width="536" alt="Screenshot 2023-05-24 at 2 58 22 PM" src="https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/97965610/2290757d-605b-4b9d-873c-d2c83f880213">

It's also already being used in some places in production. Here's an example in the claims tool:
<img width="681" alt="Screenshot 2023-05-19 at 11 48 38 AM" src="https://github.com/department-of-veterans-affairs/vets-design-system-documentation/assets/97965610/d27cc26f-04fc-40ab-b262-6e877231a05a">

## Accessibility
> Would want Platform to recommend the color and weight to comply with accessibility standards.

## Guidance
> We don't want this component to be overused where a divider is not necessary, so designers should try to design without it initially and use it if there seems to be no better way to create a hierarchy and division of content. This can be a very helpful component on content-heavy pages where a user might want to skip to another section visually if one is not of interest. We can even use this component to draw users to the beginning of a new section so important content doesn't get buried when there is a lot of information to convey.

## Research (optional)
> N/A

## Code (optional)
> N/A but would likely use the code for the existing divider and just not include the stars.

## Next steps
You may present your work to the Design System Council at an upcoming meeting. If you do not or cannot attend the Design Council Meeting, you can opt to get an asynchronous approval. 

Submit requests to join an upcoming Design System Council meeting in #platform-design-system.  

During the meeting, the Design System Council Working Group will evaluate the request and make a decision. 

[...]

PR #6153, opened 2026-04-21
Divider: Add accessibility notes to divider

babsdenney:
## Summary

Update divider component doc: add `web-component: va-divider` and anchors (Example, Usage, Accessibility considerations) in frontmatter; append an Accessibility considerations section that explains decorative vs semantic dividers and when to use `aria-hidden="true"`, with example markup for `va-divider`.

Just a note about dividers - We have an issue to add the plain divider that we should work on in the future. 

#1785 

## Related Issue

Closes #6101 

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/6153)


derekwang99 (approved):

src/_components/divider.md
```diff
@@ -24,5 +28,32 @@ mobile-app: false
...
+### Decorative vs semantic dividers
+
+Not all dividers should be announced to screen readers. Most dividers are decorative when used alongside headings or within well-structured content.
+
+**Use `aria-hidden="true"` for decorative dividers when:**
+- The divider only provides visual separation
+- Content structure is already clear through headings or other markup
+
+```html
+<va-divider aria-hidden="true"></va-divider>
```
jeana-adhoc: @babsdenney - check this - I think the divider is still just html with classes, and not an actual va-component.

jeana-adhoc: https://design.va.gov/storybook/?path=/docs/components-divider--docs

jeana-adhoc (changes requested):

src/_components/divider.md
babsdenney: Oh, you are right! I totally trusted the copilot too much on that one. Thank you for finding that. I removed the HTML examples since those aren't really necessary.

jeana-adhoc (approved):
LGTM!

PR #6164, opened 2026-04-21
Alert: Document dismissible alert persistence

### LINKED ISSUE #5709: Dismissible alert does not stay dismissed in future sessions
# Bug Report

I chose the ticket type "Bug report" because I didn't see options that were more accurate. Also this situation does fit the criteria of "the component didn't work as expected," or at least it didn't work as I expected.

- [x] I’ve searched for any related issues and avoided creating a duplicate issue.

## What happened
[In a slack thread on the platform design system channel](https://dsva.slack.com/archives/C01DBGX4P45/p1770839314554129), we learned that a dismissible alert, when dismissed, does not stay dismissed in future sessions.

## What I expected to happen
I expect and I assume users expect that when the user dismisses an alert in one session that alert will remain dismissed in future sessions. In other words, they'd expect to not experience the alert again. 

## Urgency

How urgent is this request? Please select the appropriate option below and/or provide details

- [ ] This bug is blocking work currently in progress
- [ ] This bug is affecting work currently in progress but we have a workaround
- [ ] This bug is blocking work planned within the next few sprints
- [ ] This bug is not blocking any work
- [x] Other

### Details

If a dismissible alert was available, the messages team has an active use/need for it for the OH transition. We want to convey information for a time-based period that does not persist. Our option now, especially on the app, is to have an alert that persists and eventually is no longer relevant. To solve for this in the present, the team is using the [in-product education component (also called tooltip)](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/digital-health-modernization/mhv-to-va.gov/transition/In%20Product%20Education%20Initiative%20Brief.md) to allow users to permanently dismiss the component.

babsdenney:
## Summary

Clarify in the alert component docs that dismissible alerts do not persist their dismissed state by default. Added notes to both the main alert and slim alert sections advising that applications must implement state persistence if alerts should remain dismissed across browser/app sessions.

## Related Issue

Closes #5709 

## Preview Environment Links

[Open Preview Environment](https://dev-design.va.gov/6164)


jeana-adhoc (approved):
LGTM! Today I learned that we can have dismissable slim alerts too!

